PR #10, opened 2021-06-21
CR-4769 - repo install + create git_source

ghost:


go.mod
```diff
@@ -26,6 +27,7 @@ require (
 
 replace (
 	github.com/argoproj-labs/applicationset => github.com/argoproj-labs/applicationset v0.0.0-20210614145856-2c62537a8e5a
+	github.com/argoproj-labs/argocd-autopilot => github.com/argoproj-labs/argocd-autopilot v0.0.5-0.20210621142203-8a883a7232bb
```
ghost: Temporary until we release the next autopilot

pkg/cdUtils/cdUtils.go
```diff
@@ -0,0 +1,81 @@
...
+			Labels: map[string]string{
+				"app.kubernetes.io/managed-by": store.Get().BinaryName,
+				"app.kubernetes.io/name":       opts.Name,
+			},
+			Finalizers: []string{
+				"resources-finalizer.argocd.argoproj.io",
+			},
+		},
+		Spec: argocdv1alpha1.ApplicationSpec{
+			Project: "default",
```
oren-codefresh: project should be paramertized

cmd/commands/runtime.go
```diff
@@ -320,193 +352,213 @@ func createRBAC(repofs fs.FS, path, runtimeName string) error {
...
+		Namespace: runtimeName,
+		RepoURL:   gsCloneOpts.URL(),
+		Revision:  gsCloneOpts.Revision(),
+		SrcPath:   fullGsPath,
+	})
+	if err = insFs.WriteYamls(insFs.Join(resPath, gsName+"-synchronize.yaml"), app); err != nil {
+		return err
+	}
+
+	if err = billyUtils.WriteFile(gsFs, gsFs.Join(gsPath, "DUMMY"), []byte{}, 0666); err != nil {
```
oren-codefresh: let's switch with real workflow tempalte

cmd/commands/runtime.go
```diff
@@ -320,193 +352,213 @@ func createRBAC(repofs fs.FS, path, runtimeName string) error {
...
+	err = gsRepo.Persist(ctx, &git.PushOptions{
+		CommitMsg: fmt.Sprintf("Created %s Directory", gsPath),
+	})
+	if err != nil {
+		return err
+	}
+
+	fullResPath := insFs.Join(insFs.Root(), resPath)
+	if err = createApp(ctx, insCloneOpts, runtimeName, gsName, insCloneOpts.URL()+fullResPath, application.AppTypeDirectory, runtimeName); err != nil {
+		return fmt.Errorf("failed to create git-source: %w", err)
```
oren-codefresh, line 592: What happens if some of the steps failed?

oren-codefresh (approved):

cmd/commands/runtime.go
ghost: Currently, each autopilot command (bootstrap, project create, 3 * app create + 1 final app create for the git-source) performs its own clone-persist-commit-push. So a failure in a later step will still leave the GitOps repo with some data in it, and the cluster with some state of the runtime.
We thought about using a single FS for all the commands, and change the autopilot so it will run in a mode that just uses an existing local git repo, with pull-persist-commit only, and then just manually push at the end. This way the GitOps repo will either contain all the manifests, or none.

PR #15, opened 2021-06-29
CR-5154 - wait until events is synced

ghost:


cmd/commands/runtime.go
```diff
@@ -92,120 +94,103 @@ func NewRuntimeCreateCommand() *cobra.Command {
...
 	}
 
 	return nil
 }
 
-func createApp(ctx context.Context, cloneOpts *git.CloneOptions, projectName, appName, appURL, appType, namespace string) error {
+func createApp(ctx context.Context, f kube.Factory, cloneOpts *git.CloneOptions, projectName, appName, appURL, appType, namespace string) error {
+	timeout := time.Duration(0)
+	if f != nil {
+		timeout = store.Get().WaitTimeout
```
oren-codefresh, line 191: Better to have timeout as a separate parameter and not implicit one

oren-codefresh (approved):

PR #19, opened 2021-07-12
CR-5279 - As a user I'll be able to upgrade runtime to a specific version

ghost:


pkg/runtime/runtime.go
```diff
@@ -0,0 +1,269 @@
...
+	)
+
+	if strings.HasPrefix(store.RuntimeDefURL, "http") {
+		urlObj, err := url.Parse(store.RuntimeDefURL)
+		if err != nil {
+			return nil, fmt.Errorf("failed parsing url: %w", err)
+		}
+
+		if urlObj.Query().Get("ref") == "" {
+			urlObj.Query().Set("ref", version.String())
```
oren-codefresh, line 79: should check wether version was set, it can be nil

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,269 @@
...
+		Data: map[string]string{
+			"runtime": string(data),
+		},
+	}
+
+	return fs.WriteYamls(filename, cm)
+}
+
+func (r *Runtime) Upgrade(fs fs.FS, newRt *Runtime) ([]AppDef, error) {
+	newComponents, err := r.Spec.upgrade(fs, &newRt.Spec)
```
oren-codefresh, line 149: Why upgrade only return the components to be created but do the actual delete ? I suggest that it will also do the creation/update

oren-codefresh (approved):

pkg/runtime/runtime.go
ghost: Creating a new app is a bit more complicated than just deleting it. So I preferred leaving it to autopilot.
In the autopilot, we always do clone->change->commit->push, so I had to call all the `RunAppCreate` after the commit/push of the rest of the changes (update and delete of applications)
I think we should refactor Autopilot to be able to get an existing filesystem, with an already existing repo cloned to it, and then just make the changes, without commit/push at the end. I already made a plan on how to implement it. It will make a lot of our cli-v2 code much simpler.

pkg/runtime/runtime.go
ghost: If the user doesn't supply a version, the command puts in the current binary version as the version. So when we get to here, it will always contain a valid value.

PR #29, opened 2021-07-27
CR-5747

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -248,25 +248,27 @@ func RunRuntimeList() error {
...
-		}
-		if rt.Version != nil {
-			version = *rt.Version
+		status := rt.HealthStatus
+		namespace := rt.Metadata.Namespace
+		name := rt.Metadata.Name
+		cluster := rt.Cluster
+		version := rt.Version
+
+		// TODO: understand this part of the code
```
elad-codefresh: The commented-out section is just adjusted according to the updated query, but
I need to understand what this was supposed to accomplish in order to complete the change

cmd/commands/runtime.go
```diff
@@ -248,26 +248,12 @@ func RunRuntimeList() error {
...
-		if rt.ObjectMeta != nil && rt.ObjectMeta.Name != nil {
-			name = *rt.ObjectMeta.Name
-		}
-		if rt.Cluster != nil {
-			cluster = *rt.Cluster
-		}
-		if rt.Version != nil {
-			version = *rt.Version
-		}
+		status := rt.HealthStatus
```
elad-codefresh: about rt.HealthStatus:
I do include it in the query (in go-sdk), and it is part of the Runtime model.
However, the value rt.HealthStatus holds here is nil - and it also doesn't exist on the database.
Shouldn't it exist in the database?
Is there something missing in the installation or in my understanding here?

go.mod
```diff
@@ -30,6 +30,7 @@ require (
 replace (
 	github.com/argoproj-labs/applicationset => github.com/argoproj-labs/applicationset v0.0.0-20210614145856-2c62537a8e5a
 	github.com/argoproj/argo-events => github.com/argoproj/argo-events v0.17.1-0.20210615165534-d403c441bc1d
+	github.com/codefresh-io/go-sdk => /Users/elad-codefresh/work/go-sdk
```
ghost: 👆🏻

cmd/commands/runtime.go
elad-codefresh: please notice that the current installation process does not include saving HealthStatus in the database

cmd/commands/runtime.go
```diff
@@ -250,24 +250,30 @@ func RunRuntimeList() error {
...
 		status := "N/A"
 		namespace := "N/A"
-		name := "N/A"
 		cluster := "N/A"
+		name := "N/A"
 		version := "N/A"
-		if rt.Status != nil {
-			status = rt.Status.String()
+
+		if rt.HealthMessage != nil {
```
elad-codefresh, line 257: decided to leave it N/A instead of omitting it, if that's ok

elad-codefresh:
thanks

oren-codefresh (approved):

PR #35, opened 2021-08-02
Cr 5532

daniel-codefresh:


Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION=v0.0.40
+VERSION=v0.0.41
```
ghost, line 1: please run a project-wide search-and-replace for '0.0.40' -> '0.0.41' (don't match only whole words, so it will match 0.0.40 and v0.0.40).
It should change `release_notes.md` and `runtime.yaml` as well.

daniel-codefresh: done

noam-codefresh (approved):

PR #41, opened 2021-08-03
CR-5778-2

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -274,7 +279,7 @@ func RunRuntimeList() error {
 	for _, rt := range runtimes {
 		status := "N/A"
 		namespace := "N/A"
-		cluster := "N/A"
+		cluster := rt.Cluster
 		name := "N/A"
 		version := "N/A"
```
roi-codefresh: version?

cmd/commands/runtime.go
```diff
@@ -274,7 +279,7 @@ func RunRuntimeList() error {
 	for _, rt := range runtimes {
 		status := "N/A"
 		namespace := "N/A"
-		cluster := "N/A"
```
roi-codefresh: Why is this different?

cmd/commands/runtime.go
ghost: resolved. Along with namespace and name, which are also not null

roi-codefresh (approved):

PR #50, opened 2021-08-08
CR-5108

elad-codefresh:


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ghost: I guess the two yarn files got here by mistake

cmd/commands/git-source.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+	}
+
+	appDef := &runtime.AppDef{
+		Name: opts.gsName,
+		Type: application.AppTypeDirectory,
+		URL:  opts.gsCloneOpts.Repo,
+	}
+	if err := appDef.CreateApp(ctx, nil, opts.insCloneOpts, opts.runtimeName, store.Get().CFGitSourceType, nil); err != nil {
+		return err
```
ghost:
```suggestion
		return fmt.Errorf("failed to create git-source application. Err: %w", err)
```

cmd/commands/git-source.go
```diff
@@ -0,0 +1,204 @@
...
+				gsCloneOpts.Auth.Password = insCloneOpts.Auth.Password
+			}
+
+			insCloneOpts.Parse()
+			gsCloneOpts.Parse()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+
+			err := RunCreateGitSource(ctx, &GitSourceCreateOptions{
```
ghost: just return the RunCreateGitSource function call
the errors inside print good error messages to the user

cmd/commands/git-source.go
```diff
@@ -0,0 +1,204 @@
...
+
+	log.G(ctx).Infof("done creating a new git-source: '%s'", opts.gsName)
+
+	return nil
+}
+
+func createDemoWorkflowTemplate(gsFs fs.FS, gsName, runtimeName string) error {
+	var err error
+
+	gsPath := gsFs.Join(apstore.Default.AppsDir, gsName, runtimeName)
```
ghost: the gsPath can contain the "demo-wf-template.yaml" string here, and then there is no need to call `gsFs.join` again in line 199

cmd/commands/runtime.go
```diff
@@ -258,7 +260,7 @@ func RunRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
 
 func NewRuntimeListCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "list ",
+		Use:     "list",
```
ghost:
```suggestion
		Use:     "list [runtime_name]",
```

noam-codefresh (approved):

PR #74, opened 2021-08-24
create-runtime-mutation-without-persisting

elad-codefresh:


auto-pilot972623843/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
elad-codefresh: I don't know why this got included in the pr

oren-codefresh: remove it

oren-codefresh (approved):

PR #79, opened 2021-08-26
CR-6176-list-with-version-and-cluster

elad-codefresh:


pkg/store/store.go
```diff
@@ -91,6 +92,7 @@ func init() {
 	s.ArgoWFServiceName = "argo-server"
 	s.ArgoWFServicePort = 2746
 	s.BinaryName = binaryName
+	s.Codefresh = "codefresh"
```
roi-codefresh, line 95: lol

roi-codefresh (approved):

PR #109, opened 2021-09-20
Fix logs

roi-codefresh:


cmd/commands/config.go
```diff
@@ -88,7 +88,7 @@ func RunConfigCreateContext(ctx context.Context, context, apiKey, url string) er
 	if err := cfConfig.CreateContext(ctx, context, apiKey, url); err != nil {
 		return err
 	}
-	log.G().Infof("create new context: %s", context)
+	log.G().Infof("Create new context: %s", context)
```
ghost: If the logs are after the action, shouldn't they be in past tense ("Created new context", or "New context created", or something like that)?

noam-codefresh (approved):

cmd/commands/config.go
roi-codefresh: ofc

PR #110, opened 2021-09-20
runtime-mutation-using-input-type

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -72,6 +72,14 @@ type (
 		commonConfig *runtime.CommonConfig
 	}
 
+	RuntimeCreateOnPlatformOptions struct {
```
roi-codefresh: this should come from the sdk, no defined manually

roi-codefresh (changes requested):

go.sum
```diff
@@ -266,8 +266,6 @@ github.com/clusterhq/flocker-go v0.0.0-20160920122132-2b8b7259d313/go.mod h1:P1w
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/cockroachdb/datadriven v0.0.0-20190809214429-80d97fb3cbaa/go.mod h1:zn76sxSg3SzpJ0PPJaLDCu+Bu0Lg3sKTORVIj19EIF8=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd/go.mod h1:sE/e/2PUdi/liOCUjSTXgM1o87ZssimdTWN964YiIeI=
-github.com/codefresh-io/go-sdk v0.34.5 h1:9ZPVd1bZgExRRXfttNd0XqB9CIWdDK6qRSQLpjLXIWY=
-github.com/codefresh-io/go-sdk v0.34.5/go.mod h1:CcoVmTFWHGkbrSW8LyOGB/vJe5Vzr3iC/pNE2QIBTyg=
```
elad-codefresh, line 270: after go-sdk will be merged, I will run go mod tidy again, and then (if I'm not wrong) I will not get the error I'm getting now when running that command:
go: github.com/codefresh-io/go-sdk@v0.34.6: reading github.com/codefresh-io/go-sdk/go.mod at revision v0.34.6: unknown revision v0.34.6

cmd/commands/runtime.go
```diff
@@ -204,24 +203,38 @@ func NewRuntimeInstallCommand() *cobra.Command {
 	return cmd
 }
 
-func getComponents(rt *runtime.Runtime, opts *RuntimeInstallOptions) []string {
-	var componentNames []string
+func getComponents(rt *runtime.Runtime, opts *RuntimeInstallOptions) []*string {
+	var componentNames []*string
```
roi-codefresh: you should not change this type.

cmd/commands/runtime.go
```diff
@@ -204,24 +203,38 @@ func NewRuntimeInstallCommand() *cobra.Command {
...
 
 	//  should find a more dynamic way to get these additional components
 	additionalComponents := []string{"events-reporter", "workflow-reporter"}
 	for _, additionalComponentName := range additionalComponents {
-		componentNames = append(componentNames, fmt.Sprintf("%s-%s", opts.RuntimeName, additionalComponentName))
+		componentFullName := fmt.Sprintf("%s-%s", opts.RuntimeName, additionalComponentName)
+		componentNames = append(componentNames, &componentFullName)
 	}
-	componentNames = append(componentNames, "argo-cd")
+	argoCDFullName := "argo-cd"
```
roi-codefresh: don't we have it in store?

cmd/commands/runtime.go
```diff
@@ -204,24 +203,38 @@ func NewRuntimeInstallCommand() *cobra.Command {
...
+
+	runtimeCreationResponse, err := cfConfig.NewClient().V2().Runtime().Create(ctx, &model.RuntimeInstallationArgs{
+		RuntimeName:    opts.RuntimeName,
+		Cluster:        opts.Cluster,
+		RuntimeVersion: opts.RuntimeVersion,
+		ComponentNames: opts.ComponentNames,
+		IngressHost:    opts.IngressHost,
+	})
+
+	if err != nil {
```
roi-codefresh, line 228: this is not done correctly
What if `ErrorMessage` is not empty but `err` is nil? and vice versa

cmd/commands/runtime.go
```diff
@@ -252,7 +265,13 @@ func RunRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
 
 	componentNames := getComponents(rt, opts)
 
-	token, err := createRuntimeOnPlatform(ctx, opts.RuntimeName, server, runtimeVersion, opts.IngressHost, componentNames)
+	token, err := createRuntimeOnPlatform(ctx, &model.RuntimeInstallationArgs{
+		RuntimeName:    opts.RuntimeName,
+		Cluster:         server,
+		RuntimeVersion: runtimeVersion,
+		IngressHost:    &opts.IngressHost,
```
roi-codefresh, line 261: the reason `IngressHost` is a string point is that it's optional. If it's empty you should send `nil` instead

roi-codefresh (changes requested):

cmd/commands/runtime.go
elad-codefresh: the intention is that:
* if err is nil - it means the graphql request returned with 200. so this is in order to catch bad requests.
* if ErrorMessage is not empty - it means that while the request was good, there was an error while trying to create one of the placeholder entities (runtime, appProject, components), and the ErrorMessage will be descriptive according to which step failed.

cmd/commands/runtime.go
elad-codefresh: we don't but I'll add it

cmd/commands/runtime.go
```diff
@@ -207,24 +206,34 @@ func NewRuntimeInstallCommand() *cobra.Command {
...
 	return componentNames
 }
 
-func createRuntimeOnPlatform(ctx context.Context, runtimeName string, server string, runtimeVersion string, ingressHost string, componentNames []string) (string, error) {
-	runtimeCreationResponse, err := cfConfig.NewClient().V2().Runtime().Create(ctx, runtimeName, server, runtimeVersion, ingressHost, componentNames)
+func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallationArgs) (string, error) {
 
-	if runtimeCreationResponse.ErrorMessage != nil {
-		return runtimeCreationResponse.NewAccessToken, fmt.Errorf("failed to create a new runtime: %s. Error: %w", *runtimeCreationResponse.ErrorMessage, err)
+	runtimeCreationResponse, err := cfConfig.NewClient().V2().Runtime().Create(ctx, &model.RuntimeInstallationArgs{
```
roi-codefresh: just pass in `opts`

roi-codefresh (approved):

PR #119, opened 2021-10-05
add demo github pipeline to runtime installation

daniel-codefresh:


cmd/commands/runtime.go
```diff
@@ -743,11 +743,21 @@ func createWorkflowsIngress(ctx context.Context, cloneOpts *git.CloneOptions, rt
...
 	ingress := ingressutil.CreateIngress(&ingressutil.CreateIngressOptions{
-		Name:        rt.Name + store.Get().IngressName,
-		Namespace:   rt.Namespace,
-		Path:        store.Get().IngressPath,
-		ServiceName: store.Get().ArgoWFServiceName,
-		ServicePort: store.Get().ArgoWFServicePort,
+		Name:      rt.Name + store.Get().IngressName,
+		Namespace: rt.Namespace,
+		Annotations: map[string]string{
+			"kubernetes.io/ingress.class":                  "nginx",
```
roi-codefresh, line 748: this means it only works for nginx

cmd/commands/git-source.go
```diff
@@ -570,3 +597,212 @@ func createDemoWorkflowTemplate(gsFs fs.FS, runtimeName string) error {
...
+		},
+		Data: map[string][]byte{
+			store.Get().GithubAccessTokenSecretKey: []byte(token),
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	if err = kubeFactory.Apply(context.TODO(), namespace, aputil.JoinManifests(secretYaml)); err != nil {
```
roi-codefresh: use real context here

roi-codefresh (approved):

cmd/commands/runtime.go
daniel-codefresh: this is some old code, not related to my changes actually.. the only thing i did is to move this snippet out of the CreateIngress util method (to make it more generic) to here

roi-codefresh (approved):

PR #125, opened 2021-10-07
CR-6132-uninstall-force-flag

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -432,7 +434,7 @@ func checkRuntimeCollisions(ctx context.Context, runtime string, kube kube.Facto
 		return fmt.Errorf("failed to get deployment '%s': %w", store.Get().ArgoCDServerName, err)
 	}
 
-	return fmt.Errorf("argo-cd is already installed on this cluster in namespace '%s', you need to uninstall it first", subjNamespace)
+	return fmt.Errorf("argo-cd is already installed on this cluster in namespace '%s', you can uninstall it by running '<cf> runtime uninstall %s --skip-checks --force'", subjNamespace, subjNamespace)
```
ghost: I think there is a way to get the actual current binary name, instead of `<cf>`
we are already using it in some places, here or in autopilot

noam-codefresh (approved):

PR #128, opened 2021-10-11
CR-6448-install-only-with-latest

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -253,6 +249,10 @@ func RunRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
 		runtimeVersion = rt.Spec.Version.String()
 	}
 
+	if err := preInstallationChecks(ctx, opts); err != nil {
```
ghost: This could return to where it was before, couldn't it? before the runtime.Download call

noam-codefresh (approved):

PR #138, opened 2021-10-20
Interactive-cli

rotem-codefresh:


cmd/commands/common.go
```diff
@@ -64,21 +66,176 @@ func IsValid(s string) (bool, error) {
...
 	if cloneOpts.Repo == "" {
-		runtimeData, err := cfConfig.NewClient().V2().Runtime().Get(ctx, args[0])
+		if fromAPI {
+			runtimeData, err := cfConfig.NewClient().V2().Runtime().Get(ctx, runtimeName)
+			if err != nil {
+				return fmt.Errorf("failed getting runtime repo information: %w", err)
+			}
+			if runtimeData.Repo != nil {
+				cloneOpts.Repo = *runtimeData.Repo
+				die(cmd.Flags().Set("repo", *runtimeData.Repo))
```
ghost, line 87: Wouldn't that be enough to automatically set the value on cloneOpts.Repo as well?

cmd/commands/common.go
```diff
@@ -64,21 +66,176 @@ func IsValid(s string) (bool, error) {
...
+func ensureRuntimeName(ctx context.Context, args []string, runtimeName *string) error {
+	if len(args) > 0 {
+		*runtimeName = args[0]
+	}
+
+	if *runtimeName == "" {
+		if !store.Get().Silent {
+			return getRuntimeNameFromUserSelect(ctx, runtimeName)
+		}
+		log.G(ctx).Fatal("must enter a runtime name")
```
ghost, line 120: Does this `Fatal` also exits the execution?
wouldn't just not doing anything cause cobra to report the missing flag?

cmd/commands/common.go
```diff
@@ -64,21 +66,176 @@ func IsValid(s string) (bool, error) {
...
+	}
+
+	return nil
+}
+
+func getRuntimeNameFromUserInput(runtimeName *string) error {
```
ghost, line 126: If this is not being used, it can be removed

cmd/commands/common.go
```diff
@@ -64,21 +66,176 @@ func IsValid(s string) (bool, error) {
...
+}
+
+func promptSummaryToUser(ctx context.Context, finalParameters map[string]string, description string) (bool, error) {
+	green := "\033[32m"
+	blue := "\033[34m"
+	colorReset := "\033[0m"
+	bold := "\033[1m"
+	underline := "\033[4m"
+	resetUnderline := "\033[24m"
+	resetBold := "\033[22m"
```
ghost: You can move those into `consts` at the top of the file

ghost: probably rename to uppercase, or whatever the standard in golang is

cmd/commands/common.go
ghost: oh, sorry. ignore that. I missed the usage from `runtime install`

cmd/commands/git-source.go
```diff
@@ -467,7 +467,7 @@ func NewGitSourceDeleteCommand() *cobra.Command {
 				log.G(ctx).Fatal("must enter git-source name")
 			}
 
-			err := ensureRepo(cmd, args, insCloneOpts)
+			err := ensureRepo(cmd, args[0], insCloneOpts, true)
 			if err != nil {
 				return err
 			}
```
ghost, line 475: shouldn't `git-source` commands also use `ensureGitToken`?

noam-codefresh (approved):

PR #146, opened 2021-11-01
CR-6826-kube-context

rotem-codefresh:


cmd/commands/common.go
```diff
@@ -265,6 +267,77 @@ func promptSummaryToUser(ctx context.Context, finalParameters map[string]string,
 	return false, nil
 }
 
+func getKubeContextNameFromUserSelect(cmd *cobra.Command, kubeContextName *string) error {
+	if !store.Get().Silent {
```
ghost: you can switch the condition, return `nil`, and have the rest of the func not-nested.

cmd/commands/common.go
```diff
@@ -265,6 +267,77 @@ func promptSummaryToUser(ctx context.Context, finalParameters map[string]string,
...
+			Items:     contextsNames,
+			Templates: templates,
+		}
+
+		_, result, err := prompt.Run()
+		if err != nil {
+			return fmt.Errorf("Prompt error: %w", err)
+		}
+
+		match, err := regexp.MatchString(`.+ \(current\)`, result)
```
ghost: Should we protect against the unlikely case somebody has a context he named "my (current)"?

ghost: Maybe we should just compare the `result` with `currentContext + " (current)"`, and if they are equal, just replace `result` with `currentContext`?
I'm not sure it will solve all edge cases, but it is a simpler way to do it.

ghost: What is the 1st return value of the `prompt.Run()`? If it's the selected index, that would be the simplest way to continue, wouldn't it?

cmd/commands/common.go
```diff
@@ -265,6 +267,77 @@ func promptSummaryToUser(ctx context.Context, finalParameters map[string]string,
...
+
+		die(cmd.Flags().Set("context", result))
+		*kubeContextName = result
+	}
+
+	return nil
+}
+
+func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error {
+	if !store.Get().Silent {
```
ghost: same here - switch condition, and reduce nesting

noam-codefresh (approved):

PR #161, opened 2021-11-28
Cr 7637 add default git intg

roi-codefresh:


cmd/commands/common.go
```diff
@@ -316,20 +317,22 @@ func getKubeContextNameFromUserSelect(cmd *cobra.Command, kubeContextName *strin
 
 func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error {
 	if store.Get().Silent {
-		return nil
-	}
+		if ingressHost != nil && *ingressHost != "" {
```
roi-codefresh: make field ingress required

cmd/commands/common.go
```diff
@@ -211,6 +211,7 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions) error {
 func getGitTokenFromUserInput(cmd *cobra.Command) error {
 	gitTokenPrompt := promptui.Prompt{
 		Label: "Git provider api token",
+		Mask:  '*',
```
roi-codefresh, line 214: use `*` instead of showing the token

oren-codefresh (approved):

PR #177, opened 2021-12-12
Wf-ingress-annotations

ghost:


manifests/runtime.yaml
```diff
@@ -1,26 +1,26 @@
 apiVersion: codefresh.io/v1alpha1
 kind: Runtime
 metadata:
-  name: "{{ name }}"
-  namespace: "{{ namespace }}"
+    name: '{{ name }}'
```
ghost, line 4: all the white spaces here were caused by the format-on-save I recently added. I can revert it.

roi-codefresh: no need

roi-codefresh (approved):

PR #178, opened 2021-12-13
CR-7951

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -1171,7 +1173,9 @@ func createWorkflowReporter(ctx context.Context, cloneOpts *git.CloneOptions, op
 		return err
 	}
 
-	if err := createSensor(repofs, store.Get().WorkflowReporterName, resPath, opts.RuntimeName, store.Get().WorkflowReporterName, "workflows", "data.object"); err != nil {
+	triggers := []string{"workflows", "rollouts"}
```
elad-codefresh, line 1182: as the "workflow-reporter" will from now on report rollouts events as well - do we want to rename it?

PR #180, opened 2021-12-15
CR-7598

elad-codefresh:


pkg/store/store.go
```diff
@@ -183,6 +186,8 @@ func init() {
 	s.GithubAccessTokenSecretObjectName = "autopilot-secret"
 	s.GithubAccessTokenSecretKey = "git_token"
 	s.ArgoCD = "argo-cd"
+	s.MinimumMemorySizeRequired = "3000"
```
elad-codefresh: what should we set as the minimum requirements?

cmd/commands/runtime.go
```diff
@@ -292,6 +307,172 @@ func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallatio
...
+	
+	client, err := kubeFactory.KubernetesClientSet()
+	if err != nil {
+		return fmt.Errorf(fmt.Sprintf("Cannot create kubernetes clientset: %v ", err))
+	}
+
+	req := validationRequest{
+		rbac: []rbacValidation{
+			{
+				Resource:  "ServiceAccount",
```
elad-codefresh: wondering if in this case I should change the hard-coded verb values and resource names to constants defined in the store

cmd/commands/runtime.go
```diff
@@ -292,6 +307,172 @@ func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallatio
 	return runtimeCreationResponse.NewAccessToken, nil
 }
 
+func ensureClusterRequirements(ctx context.Context, kubeFactory kube.Factory, namespace string) error {
+	requirementsValidationErrorMessage := "cluster does not meet minimum requirements"
+	var specificErrorMessages []string
+	
+	client, err := kubeFactory.KubernetesClientSet()
+	if err != nil {
+		return fmt.Errorf("%s: cannot create kubernetes clientset: %v ", requirementsValidationErrorMessage, err)
```
roi-codefresh: no need for the `cluster does not meet minimum requirements` here. maybe just `failed to verify cluster requirements: cannot create kubernetes clientset: %w`...

cmd/commands/runtime.go
roi-codefresh: no need

cmd/commands/runtime.go
```diff
@@ -292,6 +307,172 @@ func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallatio
...
+				Namespace: namespace,
+			},
+			{
+				Resource:  "Service",
+				Verbs:     []string{"create", "update", "delete"},
+				Namespace: namespace,
+			},
+			{
+				Resource:  "Role",
+				Group:     store.Get().RoleApiGroup,
```
roi-codefresh: no need to save it to the store

cmd/commands/runtime.go
```diff
@@ -292,6 +307,172 @@ func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallatio
...
+		}
+		return fmt.Errorf("%s: failed testing rbac: %v", requirementsValidationErrorMessage, specificErrorMessages)
+	}
+
+	nodes, err := client.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("%s: failed getting nodes: %v", requirementsValidationErrorMessage, err)
+	}
+
+	if nodes == nil {
```
roi-codefresh: don't think there's a need for this check

cmd/commands/runtime.go
```diff
@@ -292,6 +307,172 @@ func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallatio
 	return runtimeCreationResponse.NewAccessToken, nil
 }
 
+func ensureClusterRequirements(ctx context.Context, kubeFactory kube.Factory, namespace string) error {
```
roi-codefresh: please move this function and everything related to it to pkg/util/kube/kube.go

cmd/commands/runtime.go
```diff
@@ -47,6 +47,7 @@ import (
 	argowf "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow"
 
 	"github.com/Masterminds/semver/v3"
+	cliv2kube "github.com/codefresh-io/cli-v2/pkg/util/kube"
```
elad-codefresh: tested locally with 'replace'

pkg/store/store.go
```diff
@@ -183,6 +186,8 @@ func init() {
...
```
elad-codefresh, line 189: are we ok with these minimum requirements?

cmd/commands/runtime.go
roi-codefresh: name it as `kubeutil` i think it's a better name

roi-codefresh (approved):

PR #201, opened 2022-01-12
CR-8437 rollback installation in case of failure

rotem-codefresh:
+ informative summary

### DIFF
--- a/cmd/commands/common.go
+++ b/cmd/commands/common.go
@@ -360,3 +360,8 @@ func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error 
 
 	return nil
 }
+
+func decorateErrorWithDocsLink(err error) error {
+	newMsg := fmt.Sprintf("%s\nfor more information: %s", err.Error(), store.Get().DocsLink)
+	return fmt.Errorf("%s" ,newMsg)
+}
--- a/cmd/commands/common.go
+++ b/cmd/commands/common.go
@@ -361,19 +361,3 @@ func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error 
 
 	return nil
 }
-
-func verifyCLILatestVersion(ctx context.Context) error {
-	latestVersionString, err := cfConfig.NewClient().V2().CliReleases().GetLatest(ctx)
-	if err != nil {
-		return fmt.Errorf("failed getting latest cli release: %w", err)
-	}
-
-	latestVersionSemver := semver.MustParse(latestVersionString)
-	currentVersion := store.Get().Version.Version
-
-	if currentVersion.LessThan(latestVersionSemver) {
-		return fmt.Errorf("please upgrade to the latest cli version: %s", latestVersionString)
-	}
-
-	return nil
-}
--- a/cmd/commands/common.go
+++ b/cmd/commands/common.go
@@ -60,6 +60,7 @@ var (
 	cfConfig *config.Config
 
 	GREEN           = "\033[32m"
+	RED             = "\033[31m"
 	CYAN            = "\033[36m"
 	BOLD            = "\033[1m"
 	UNDERLINE       = "\033[4m"
--- a/cmd/commands/common.go
+++ b/cmd/commands/common.go
@@ -360,8 +360,3 @@ func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error 
 
 	return nil
 }
-
-func decorateErrorWithDocsLink(err error) error {
-	newMsg := fmt.Sprintf("%s\nfor more information: %s", err.Error(), store.Get().DocsLink)
-	return fmt.Errorf("%s" ,newMsg)
-}
--- a/cmd/commands/common.go
+++ b/cmd/commands/common.go
@@ -36,7 +36,6 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/argoproj-labs/argocd-autopilot/pkg/git"
 	"github.com/codefresh-io/cli-v2/pkg/config"
 	"github.com/codefresh-io/cli-v2/pkg/log"
--- a/cmd/commands/runtime.go
+++ b/cmd/commands/runtime.go
@@ -228,10 +228,6 @@ func NewRuntimeInstallCommand() *cobra.Command {
 }
 
 func runtimeInstallCommandPreRunHandler(cmd *cobra.Command, opts *RuntimeInstallOptions) error {
-	if err := verifyCLILatestVersion(cmd.Context()); err != nil {
-		return err
-	}
-
 	if opts.versionStr != "" {
 		version, err := semver.NewVersion(opts.versionStr)
 		if err != nil {
@@ -661,11 +657,6 @@ func NewRuntimeListCommand() *cobra.Command {
 		Short:   "List all Codefresh runtimes",
 		Example: util.Doc(`<BIN> runtime list`),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			ctx := cmd.Context()
-
-			if err := verifyCLILatestVersion(ctx); err != nil {
-				return fmt.Errorf("%w", err)
-			}
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, _ []string) error {
@@ -776,10 +767,6 @@ func NewRuntimeUninstallCommand() *cobra.Command {
 `),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
-
-			if err := verifyCLILatestVersion(ctx); err != nil {
-				return fmt.Errorf("%w", err)
-			}
 
 			err := getKubeContextNameFromUserSelect(cmd, &kubeContextName)
 			if err != nil {
@@ -926,10 +913,6 @@ func NewRuntimeUpgradeCommand() *cobra.Command {
 `),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
-
-			if err := verifyCLILatestVersion(ctx); err != nil {
-				return fmt.Errorf("%w", err)
-			}
 
 			err := ensureRuntimeName(ctx, args, &runtimeName)
 			if err != nil {
--- a/cmd/commands/runtime.go
+++ b/cmd/commands/runtime.go
@@ -116,6 +116,18 @@ type (
 		version      string
 		saName       string
 	}
+
+	summaryLogLevels string
+	summaryLog struct {
+		message string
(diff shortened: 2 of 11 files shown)


cmd/commands/runtime.go
```diff
@@ -314,27 +324,39 @@ func createRuntimeOnPlatform(ctx context.Context, opts *model.RuntimeInstallatio
...
 		return fmt.Errorf("pre installation checks failed: %w", err)
 	}
 
 	rt, err := runtime.Download(opts.Version, opts.RuntimeName)
+	appendLogToSummary("Dowmloading runtime definition", err)
 	if err != nil {
 		return fmt.Errorf("failed to download runtime definition: %w", err)
 	}
 
+	if rt.Spec.DefVersion.GreaterThan(store.Get().MaxDefVersion) {
```
roi-codefresh: why is this not a precheck?

cmd/commands/runtime.go
```diff
@@ -343,19 +365,16 @@ func RunRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
 		ComponentNames: componentNames,
 		Repo:           &opts.InsCloneOpts.Repo,
 	})
+	appendLogToSummary("Creating runtime on platform", err)
 	if err != nil {
-		return fmt.Errorf("failed to create a new runtime: %w", err)
+		return fmt.Errorf("failed to create a new runtime: %w\nfor more information: https://codefresh.io/csdp-docs/", err)
```
roi-codefresh: the link to the docs should probably be in the store and not hardcoded here.

cmd/commands/runtime.go
```diff
@@ -380,71 +399,85 @@ func RunRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
...
 		CreateDemoResources: false,
 		Exclude:             "**/images/**/*",
 		Include:             "workflows/**/*.yaml",
-	}); err != nil {
-		installationErr = fmt.Errorf("failed to create `%s`: %w", store.Get().MarketplaceGitSourceName, err)
-		return installationErr
+	})
+	appendLogToSummary(createGitSrcMessgae, err)
+	if err != nil {
+		return fmt.Errorf("failed to create `%s`: %w\nfor more information: https://codefresh.io/csdp-docs/", store.Get().MarketplaceGitSourceName, err)
```
roi-codefresh: this 'for more information' string should really be in some utility function that decorates errors and not repeated everywhere

cmd/commands/runtime.go
```diff
@@ -380,71 +399,85 @@ func RunRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
...
+	if timeoutErr != nil {
+		return fmt.Errorf("failed to complete installation: %w\nfor more information: https://codefresh.io/csdp-docs/", timeoutErr)
 	}
 
-	if err := addDefaultGitIntegration(ctx, opts.RuntimeName, opts.GitIntegrationOpts); err != nil {
-		return fmt.Errorf("failed to create default git integration: %w", err)
+	gitIntgErr := addDefaultGitIntegration(ctx, opts.RuntimeName, opts.GitIntegrationOpts)
+	appendLogToSummary("Creating a default git integration", gitIntgErr)
+	if gitIntgErr != nil {
+		return fmt.Errorf("failed to create default git integration: %w\nyou can create it manually with the command: `cf integration git add default --runtime %s --api-url https://github.com`", gitIntgErr, opts.RuntimeName)
```
roi-codefresh: this is not the api url of github, and we can't assume the user is using github here anyway..
bottom line, I don't think we should print this message at all.

cmd/commands/runtime.go
```diff
@@ -1494,3 +1503,43 @@ func inferAPIURLForGitProvider(provider apmodel.GitProviders) (string, error) {
 
 	return "", fmt.Errorf("cannot infer api-url for git provider %s, %s", provider, suggest)
 }
+
+func postInstallationHandler(ctx context.Context, opts *RuntimeInstallOptions, err *error) {
+	if *err != nil {
+		summaryArr = append(summaryArr, summaryLog{"----------Uninstalling runtime----------", Info})
```
roi-codefresh, line 1514: didn't we say it would be opt in?

cmd/commands/runtime.go
```diff
@@ -1494,3 +1503,43 @@ func inferAPIURLForGitProvider(provider apmodel.GitProviders) (string, error) {
...
+		summaryArr = append(summaryArr, summaryLog{message, Failed})
+	} else {
+		summaryArr = append(summaryArr, summaryLog{message, Success})
+	}
+}
+
+func printSummaryToUser() {
+	for i := 0; i < len(summaryArr); i++ {
+		if summaryArr[i].level == Success {
+			fmt.Printf("%s -> %v%s%v\n", summaryArr[i].message, GREEN, summaryArr[i].level, COLOR_RESET)
```
roi-codefresh, line 1544: usually the level should come before the message

cmd/commands/runtime.go
```diff
@@ -1494,3 +1503,43 @@ func inferAPIURLForGitProvider(provider apmodel.GitProviders) (string, error) {
...
+func appendLogToSummary(message string, err error){
+	if err != nil {
+		summaryArr = append(summaryArr, summaryLog{message, Failed})
+	} else {
+		summaryArr = append(summaryArr, summaryLog{message, Success})
+	}
+}
+
+func printSummaryToUser() {
+	for i := 0; i < len(summaryArr); i++ {
```
roi-codefresh, line 1542: does the user need to see all this again even if the installation was successful? didn't he already see it with the regular logs we had before?

cmd/commands/runtime.go
rotem-codefresh: there is a different task to add a flag. for now, it is default, except the case of a timeout or creating git integration fails

cmd/commands/runtime.go
rotem-codefresh: this is the way Idan wanted it

cmd/commands/common.go
```diff
@@ -361,18 +361,7 @@ func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error
...
-
-	latestVersionSemver := semver.MustParse(latestVersionString)
-	currentVersion := store.Get().Version.Version
-
-	if currentVersion.LessThan(latestVersionSemver) {
-		return fmt.Errorf("please upgrade to the latest cli version: %s", latestVersionString)
-	}
-
-	return nil
+func decorateErrorWithDocsLink(err error) error {
```
roi-codefresh: can we move this to `pkg/util` please?

cmd/commands/common.go
```diff
@@ -361,18 +361,7 @@ func getIngressHostFromUserInput(cmd *cobra.Command, ingressHost *string) error
...
+	newMsg := fmt.Sprintf("%s\nfor more information: %s", err.Error(), store.Get().DocsLink)
```
roi-codefresh: `Errorf` can actually format, so no need to use `Sprintf` here as well.

cmd/commands/runtime.go
```diff
@@ -540,6 +549,16 @@ func installComponents(ctx context.Context, opts *RuntimeInstallOptions, rt *run
 func preInstallationChecks(ctx context.Context, opts *RuntimeInstallOptions) error {
 	log.G(ctx).Debug("running pre-installation checks...")
 
+	rt, err := runtime.Download(opts.Version, opts.RuntimeName)
+	appendLogToSummary("Dowmloading runtime definition", err)
```
roi-codefresh:
```suggestion
	appendLogToSummary("Downloading runtime definition", err)
```

cmd/commands/runtime.go
```diff
@@ -819,40 +829,48 @@ func NewRuntimeUninstallCommand() *cobra.Command {
...
 func RunRuntimeUninstall(ctx context.Context, opts *RuntimeUninstallOptions) error {
+	defer printSummaryToUser()
+
 	// check whether the runtime exists
 	if !opts.SkipChecks {
 		_, err := cfConfig.NewClient().V2().Runtime().Get(ctx, opts.RuntimeName)
+		appendLogToSummary("Checking if runtime exists", err)
 		if err != nil {
-			log.G(ctx).Warn("you can attempt to uninstall again with the \"--skip-checks\" flag")
+			summaryArr = append(summaryArr, summaryLog{"you can attempt to uninstall again with the \"--skip-checks\" flag", Info})
```
roi-codefresh, line 839: why do we append twice here?

cmd/commands/runtime.go
```diff
@@ -819,40 +829,48 @@ func NewRuntimeUninstallCommand() *cobra.Command {
...
 		KubeFactory:  opts.KubeFactory,
 		Force:        opts.Force,
 		FastExit:     opts.FastExit,
-	}); err != nil {
+	})
+	appendLogToSummary("Uninstalling repo", err)
+	if err != nil {
 		if !opts.Force {
-			log.G(ctx).Warn("you can attempt to uninstall again with the \"--force\" flag")
+			summaryArr = append(summaryArr, summaryLog{"you can attempt to uninstall again with the \"--force\" flag", Info})
```
roi-codefresh, line 857: and here?

cmd/commands/runtime.go
```diff
@@ -819,40 +829,48 @@ func NewRuntimeUninstallCommand() *cobra.Command {
...
+	if err != nil {
 		return fmt.Errorf("failed to delete runtime from the platform: %w", err)
 	}
 
 	if cfConfig.GetCurrentContext().DefaultRuntime == opts.RuntimeName {
 		cfConfig.GetCurrentContext().DefaultRuntime = ""
 	}
 
-	log.G(ctx).Infof("Done uninstalling runtime '%s'", opts.RuntimeName)
+	installationDoneStr := fmt.Sprintf("Done uninstalling runtime '%s'", opts.RuntimeName)
```
roi-codefresh: the variable name is wrong

cmd/commands/runtime.go
roi-codefresh, line 857: why do we somtimes use this to append
```
summaryArr = append(summaryArr, summaryLog{"...
```
and sometimes this:
```
appendLogToSummary("Uninstalling repo", err)
```
?

rotem-codefresh: because the appendLogToSummary is appending with `Success` or `Failure` according to the err, and if it's `Info` level, I append it with append.

cmd/commands/runtime.go
rotem-codefresh: what's wrong with it?

roi-codefresh: it's not installation, it's uninstallation

noam-codefresh (approved):

PR #210, opened 2022-01-17
support for example pipeline

oren-codefresh:


cmd/commands/git-source.go
```diff
@@ -697,6 +700,7 @@ func createGithubExampleIngress() *netv1.Ingress {
 				ServicePort: store.Get().GithubExampleEventSourceServicePort,
 			},
 		},
+		Annotations: map[string]string{"kubernetes.io/ingress.class": "nginx"},
```
ghost: do we need that, when we have the default annotation on the IngressClass resource?
currently, our ingresses don't have anything "nginx" specific (even though I guess their entire setup is dependent upon nginx-style set up)

oren-codefresh: It's a good comment , I don't want to have the default annotation on the ingress class.. on the other hand with this one we are only using nginx.. I'll check it @noam-codefresh

noam-codefresh (approved):

roi-codefresh (approved):

PR #211, opened 2022-01-18
CR-8833

elad-codefresh:


pkg/store/store.go
```diff
@@ -213,14 +217,15 @@ func init() {
...
 	s.ReplicaSetReporterServiceAccount = "replicaset-reporter-sa"
-	s.ReplicaSetResourceName = "replicasets"
 	s.WorkflowResourceName = "workflows"
 	s.RolloutReporterName = "rollout-reporter"
-	s.RolloutResourceName = "rollouts"
 	s.RolloutReporterServiceAccount = "rollout-reporter-sa"
 	s.SegmentWriteKey = segmentWriteKey
 	s.RequirementsLink = "https://codefresh.io/csdp-docs/docs/runtime/requirements/"
 	s.DownloadCliLink = "https://codefresh.io/csdp-docs/docs/clients/csdp-cli/"
+	s.RolloutResourcesNames = []string{"rollouts", "replicasets", "analysisruns"}
```
roi-codefresh: this is so confusing 🤧

cmd/commands/runtime.go
```diff
@@ -48,7 +48,7 @@ import (
 	apstore "github.com/argoproj-labs/argocd-autopilot/pkg/store"
 	aputil "github.com/argoproj-labs/argocd-autopilot/pkg/util"
 	argocdv1alpha1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
-	argowf "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow"
+	"github.com/argoproj/argo-events/pkg/apis/eventsource/v1alpha1"
```
elad-codefresh: argowf is no more needed here, cause no longer using argowf.Group and argowf.Version, a

cmd/commands/runtime.go
```diff
@@ -723,30 +728,38 @@ func installComponents(ctx context.Context, opts *RuntimeInstallOptions, rt *run
...
-	}
-
 	if err = createReporter(ctx, opts.InsCloneOpts, opts, reporterCreateOptions{
 		reporterName: store.Get().RolloutReporterName,
-		resourceName: store.Get().RolloutResourceName,
-		group:        "argoproj.io",
-		version:      "v1alpha1",
-		saName:       store.Get().RolloutReporterServiceAccount,
+		gvk: []Gvk{
+			{
```
elad-codefresh, line 807: do you think it's needed in this case to replace these hard coded strings with values defined in the store?

pkg/store/store.go
```diff
@@ -129,7 +129,11 @@ type Store struct {
 	RolloutReporterName                  string
 	RolloutResourceName                  string
 	RolloutReporterServiceAccount        string
+	AnalysisRunResourceName              string
 	SegmentWriteKey                      string
+	RolloutResourcesNames                []string
+	RolloutResourcesGroupNames           []string
+	RolloutResourcesVersionNames         []string
```
ghost: are these needed?

cmd/commands/runtime.go
ghost: ideally, we will get these values from the matching argo-rollouts/argo-workflows/argo-whatever types.go
i don't remember if i looked for them, and couldn't find them, or we just decided at the end to not add a ton of import statements and dependencies, just to get a few strings.
i don't think that every string needs to be in the `store.go` file, personally.

noam-codefresh (approved):

roi-codefresh (approved):

roi-codefresh (approved):

PR #220, opened 2022-01-26
Added reporting for each installation/uninstallation step

danielm-codefresh:


cmd/commands/runtime.go
```diff
@@ -193,7 +196,7 @@ func NewRuntimeInstallCommand() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return RunRuntimeInstall(cmd.Context(), &installationOpts)
+			return RunRuntimeInstall(cmd.Context(), &installationOpts, reporter.Get())
```
roi-codefresh: no need to pass it here anymore

cmd/commands/runtime.go
```diff
@@ -851,13 +856,22 @@ func NewRuntimeUninstallCommand() *cobra.Command {
 	return cmd
 }
 
-func RunRuntimeUninstall(ctx context.Context, opts *RuntimeUninstallOptions) error {
+func createAnalyticsReporter(ctx context.Context) {
+	user, err := cfConfig.GetCurrentContext().GetUser(ctx)
+	// If error, it will default to noop reporter
+	if err != nil {
+		return
```
roi-codefresh: print debug message here, later I will figure out why we can see debug messages.
but you should not swallow the error here.

cmd/commands/runtime.go
```diff
@@ -836,7 +841,7 @@ func NewRuntimeUninstallCommand() *cobra.Command {
 				SkipChecks:  skipChecks,
 				Force:       force,
 				FastExit:    fastExit,
-			})
+			}, reporter.Get())
```
roi-codefresh: should no longer be passed here.

cmd/commands/runtime.go
```diff
@@ -789,9 +792,12 @@ func NewRuntimeUninstallCommand() *cobra.Command {
 	<BIN> runtime uninstall runtime-name --repo gitops_repo
 `),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			var err error
 			ctx := cmd.Context()
 
-			err := getKubeContextNameFromUserSelect(cmd, &kubeContextName)
+			createAnalyticsReporter(ctx)
+
+			err = getKubeContextNameFromUserSelect(cmd, &kubeContextName)
```
roi-codefresh, line 801: this should be inside the `if` statement

cmd/commands/runtime.go
```diff
@@ -851,13 +856,22 @@ func NewRuntimeUninstallCommand() *cobra.Command {
 	return cmd
 }
 
-func RunRuntimeUninstall(ctx context.Context, opts *RuntimeUninstallOptions) error {
+func createAnalyticsReporter(ctx context.Context) {
```
roi-codefresh: this function should be at the bottom of the file, not between run methods.

cmd/commands/runtime.go
```diff
@@ -1554,6 +1568,21 @@ func postInstallationHandler(ctx context.Context, opts *RuntimeInstallOptions, e
 	printSummaryToUser()
 }
 
+func HandleCliStep(ctx context.Context, r reporter.AnalyticsReporter, event reporter.CliEventType, message string, err error) {
```
roi-codefresh: This should be a private function inside this package. not public.

roi-codefresh: also `reporter` should not be passed here (use `.Get()`).

roi-codefresh: also `ctx` is not used here

roi-codefresh: also you can change this to use the `...error` as last parameter, then it won't be required to pass `nil` if there is no error.

cmd/commands/runtime.go
```diff
@@ -1554,6 +1568,21 @@ func postInstallationHandler(ctx context.Context, opts *RuntimeInstallOptions, e
...
+	status := reporter.SUCCESS
+	appendLogToSummary(message, err)
```
roi-codefresh: this should not sit between the `status` declaration and the `if`. maybe move it to the beginning of this function.

pkg/config/config.go
```diff
@@ -320,3 +326,7 @@ func init() {
 	}
 	defaultPath = homedir
 }
+
+func (a *AuthContext) GetUser(ctx context.Context) (*codefresh.User, error) {
```
roi-codefresh, line 330: maybe cover with some UTs here.

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	CliStepData struct {
+		Event       CliEventType
+		Status      CliStepStatus
+		Description string
+		Err         error
+	}
+
+	segmentAnalyticsReporter struct {
```
roi-codefresh, line 43: move public types to the top, private types to the bottom

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
+package reporter
+
+import (
+	"github.com/codefresh-io/cli-v2/pkg/log"
+	"github.com/codefresh-io/cli-v2/pkg/store"
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
+var ar AnalyticsReporter
```
roi-codefresh: you can already initialize to `noopAnalyticsReporter`

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
...
+
+	noopAnalyticsReporter struct{}
+
+	CliEventType  string
+	CliStepStatus string
+)
+
+const (
+	// Install
+	PRE_INSTALLATION_CHECKS        CliEventType = "install.pre-installation-checks"
```
roi-codefresh: Please follow the golang coding convention when naming consts:
https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const

roi-codefresh: Also, think of having prefix inside the name to indicate if it's an install or uninstall step, because they have the same type it might be confusing if there are similar steps.
For example: `InstallStepPreChecks`, `InstallStepDownloadRuntimeDefinition`
And: `UninstallStepCheckRuntimeExists`....

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
...
+	CANCELED CliStepStatus = "CANCELED"
+)
+
+// Get returns the global reporter
+func Get() AnalyticsReporter {
+	return ar
+}
+
+func init() {
+	ar = &noopAnalyticsReporter{}
```
roi-codefresh: this can be moved to inline and avoided

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func CreateAnalyticsReporter(userId, accountId string) {
```
roi-codefresh: I would name it `Init()`. it looks cooler.

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
...
+	writeKey := store.Get().SegmentWriteKey
+	if writeKey == "" {
+		log.G().Debug("No segment write key was provided. Using the noop reporter.")
```
roi-codefresh, line 93: you need to return here, otherwise it's a bug.

pkg/util/util.go
```diff
@@ -53,6 +54,12 @@ func ContextWithCancelOnSignals(ctx context.Context, sigs ...os.Signal) context.
 		for {
 			s := <-sig
 			cancels++
+			reporter.Get().ReportStep(reporter.CliStepData{
```
roi-codefresh: actually, now that I think about it, there is no way from watching the events received, to connect them into a single installation attempt. This means it would hard to connect back the dots and see that a specific installation attempt was ended in a cancellation. 
To solve it I would generate inside the `Init` method of the analytics reporter a uid that would be reported for each event and called `flowId`. later this id can be used to trace back a whole flow, or installation attempt.

roi-codefresh (approved):

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,106 @@
+package reporter
+
+import (
+	"github.com/codefresh-io/cli-v2/pkg/log"
+	"github.com/codefresh-io/cli-v2/pkg/store"
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
```
roi-codefresh, line 24: I would use `var G = Get` here, just because it's cooler to use `ar.G().ReportStep...` instead of `ar.Get()...`

roi-codefresh (approved):

PR #232, opened 2022-01-31
CR-8938 verify github token's scopes

rotem-codefresh:


pkg/git/github.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/codefresh-io/cli-v2/pkg/log"
+)
+
+func VerifyGitHubTokenScope(ctx context.Context, token string) (bool, error) {
+	log.G(ctx).Info("Verifing your git token")
+
+	req, _ := http.NewRequest("HEAD", "https://api.github.com/", nil)
+	req.Header.Set("Authorization", fmt.Sprintf("token %s", token))
+
+	resp, err := http.DefaultClient.Do(req)
```
roi-codefresh: you need to use `ctx` when sending the request.

pkg/git/github.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	defer resp.Body.Close()
+
+	rawScopes := resp.Header["X-Oauth-Scopes"]
+	var scopes []string
+	if len(rawScopes) > 0 {
+		scopes = strings.Split(rawScopes[0], ", ")
+	}
+	var repo bool
```
roi-codefresh: save the required scopes in an array var at the top of the file

pkg/git/gitlab.go
```diff
@@ -0,0 +1,13 @@
+package git
+
+import (
+	"context"
+
+	"github.com/codefresh-io/cli-v2/pkg/log"
+)
+
+func VerifyGitLabTokenScope(ctx context.Context, token string) (bool, error) {
```
roi-codefresh: If the function name includes the provider, I don't see a point in separating it to different files

pkg/git/index.go
```diff
@@ -0,0 +1,15 @@
+package git
+
+import "context"
+
+func GetTokenVerifier(provider string) func(context.Context, string)(bool, error) {
```
roi-codefresh: there is no `index.go` files in go. This should be called `git.go`

roi-codefresh: This function should be just called `VerifyToken(token, provider)`

pkg/git/index.go
```diff
@@ -0,0 +1,15 @@
...
+	if provider == "github" {
+		return VerifyGitHubTokenScope
```
roi-codefresh: you should have a `map[string]func..` instead

pkg/util/util.go
```diff
@@ -171,8 +171,8 @@ func CurrentServer() (string, error) {
 	return server, nil
 }
 
-func DecorateErrorWithDocsLink(err error) error {
-	return fmt.Errorf("%s\nfor more information: %s", err.Error(), store.Get().DocsLink)
+func DecorateErrorWithDocsLink(err error, link string) error {
```
roi-codefresh: instead of breaking all calls to this function you should have made this optional by using `(err error, link ...string)`. Then you can pass the link arg only when needed

rotem-codefresh: but without link, this function is not decorating anything!

pkg/git/github.go
rotem-codefresh: please explain

roi-codefresh: https://pkg.go.dev/net/http#NewRequestWithContext

pkg/util/util.go
roi-codefresh: it is, `store.Get().DocsLink` would be the default.

pkg/git/git.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+
+var (
+	requiredGitHubScopes = []string{ "repo", "admin:repo_hook" }
+)
+
+
+func VerifyToken(ctx context.Context, provider string, token string) (bool, error) {
+	providerToVerifier := map[string]func(context.Context, string)(bool, error){}
```
roi-codefresh: why not
```golang
	providerToVerifier := map[string]func(context.Context, string)(bool, error){
                "github": verifyGitHubTokenScope,
                "gitlab": verifyGitLabTokenScope,
        }
```

pkg/git/git.go
```diff
@@ -0,0 +1,70 @@
...
+
+	providerToVerifier["github"] = verifyGitHubTokenScope
+	providerToVerifier["gitlab"] = verifyGitLabTokenScope
+
+	verifier := providerToVerifier[provider]
```
roi-codefresh, line 37: what if `verifier` is `nil`?

pkg/git/git.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return verifier(ctx, token)
+}
+
+func verifyGitHubTokenScope(ctx context.Context, token string) (bool, error) {
+	log.G(ctx).Info("Verifing your git token")
```
roi-codefresh: This log can be outside, possibly in `runtime.go`

pkg/git/git.go
```diff
@@ -0,0 +1,70 @@
...
+	rawScopes := resp.Header["X-Oauth-Scopes"]
+	var scopes []string
+	if len(rawScopes) > 0 {
+		scopes = strings.Split(rawScopes[0], ", ")
+	}
+	var repo bool
+	var adminRepoHook bool
+
+	for _, scope := range scopes {
+		if scope == requiredGitHubScopes[0] {
```
roi-codefresh: This is not what I expected..
you have an array of scopes you need to verify. tomorrow I will add additional 10 scopes.

pkg/git/git.go
rotem-codefresh: it can't be, the installation will fail if the provider is not github or gitlab

pkg/git/git.go
rotem-codefresh: i thought so too, but it will look weird in case of gitlab provider

pkg/git/git.go
roi-codefresh: and you think if tomorrow someone adds bitbucket they will remember to add it here?

pkg/git/git.go
roi-codefresh: why?

roi-codefresh: so maybe just remove it, library code should generally not print info logs.

pkg/git/git.go
roi-codefresh: I think you should only have github here, and any other provider for now should go to a fallback verifier that looks something like the gitlab one.

cmd/commands/common.go
```diff
@@ -247,11 +249,37 @@ func getIngressClassFromUserSelect(ctx context.Context, ingressClassNames []stri
...
+	verified, err := cfgit.VerifyToken(cmd.Context(), cloneOpts.Provider, cloneOpts.Auth.Password)
+	if err != nil {
+		return fmt.Errorf("failed to verify git token: %w", err)
+	}
+
+	if verified {
+		return nil
+	}
+	
+	return fmt.Errorf("The provided git token is missing one or more of the required scopes")
```
roi-codefresh: I think it could be huge if we actually print the scopes here. What I don't understand though is why it should even be here. Why don't the verifier return only error instead of bool and error. And the error can be formatted in each verifier.

cmd/commands/runtime.go
```diff
@@ -287,6 +287,12 @@ func runtimeInstallCommandPreRunHandler(cmd *cobra.Command, opts *RuntimeInstall
 		return err
 	}
 
+	opts.InsCloneOpts.Provider, err = inferProviderFromRepo(opts.InsCloneOpts.Repo)
```
roi-codefresh: why do we infer provider always? what if --provider is used?

rotem-codefresh: i added a function that will infer the provider from the repo in an early stage, and if it can't it will fail. so this will always be with a valid value

roi-codefresh: I don't understand. what if i use `cf runtime install foobar --provider github --repo https://some-git-hub.com/foo/bar`. Why is that not okay?

cmd/commands/common.go
```diff
@@ -247,10 +249,34 @@ func getIngressClassFromUserSelect(ctx context.Context, ingressClassNames []stri
...
+			return err
+		}
 	}
+
+	err := cfgit.VerifyToken(cmd.Context(), cloneOpts.Provider, cloneOpts.Auth.Password)
+	if err != nil {
+		return fmt.Errorf("failed to verify git token: %w", err)
+	}
+
+	log.G(cmd.Context()).Info("Token varified")
```
```suggestion
	log.G(cmd.Context()).Info("Token verified")
```

cmd/commands/runtime.go
```diff
@@ -439,11 +441,11 @@ func ensureIngressClass(ctx context.Context, opts *RuntimeInstallOptions) error
 			opts.IngressController = ingressClassNameToController[opts.IngressClass]
 			return nil
 		}
-		return fmt.Errorf("Ingress Class '%s' is not supported. Only the ingress class of type NGINX is supported. for more information: %s", opts.IngressClass, store.Get().RequirementsLink)
+		return util.DecorateErrorWithDocsLink(fmt.Errorf("Ingress Class '%s' is not supported. Only the ingress class of type NGINX is supported.", opts.IngressClass), store.Get().RequirementsLink)
```
roi-codefresh: Errors should be all lowercase

cmd/commands/runtime.go
```diff
@@ -1727,6 +1732,17 @@ func inferProviderFromCloneURL(cloneURL string) (apmodel.GitProviders, error) {
 	return apmodel.GitProviders(""), fmt.Errorf("failed to infer git provider from clone url: %s, %s", cloneURL, suggest)
 }
 
+func inferProviderFromInsProvider(provider string) apmodel.GitProviders {
+	if provider == "github" {
```
roi-codefresh: you can get rid of this function and do `apmodel.GitProviders(strings.ToUpper(provider))`

pkg/git/git.go
```diff
@@ -0,0 +1,86 @@
...
+		return verifyTokenScopeFallback(ctx, provider)
+	}
+
+	return verifier(ctx, token)
+}
+
+func verifyGitHubTokenScope(ctx context.Context, token string) error {
+	errMessage := "The provided git token is missing one or more of the required scopes:" + util.StringifyArray(requiredGitHubScopes)
+	
+	req, _ := http.NewRequestWithContext(ctx, "HEAD", "https://api.github.com/", nil)
```
roi-codefresh: why do you ignore the error here?

pkg/git/git.go
```diff
@@ -0,0 +1,86 @@
...
+	verifier := providerToVerifier[provider]
+	if verifier == nil {
+		return verifyTokenScopeFallback(ctx, provider)
+	}
+
+	return verifier(ctx, token)
+}
+
+func verifyGitHubTokenScope(ctx context.Context, token string) error {
+	errMessage := "The provided git token is missing one or more of the required scopes:" + util.StringifyArray(requiredGitHubScopes)
```
roi-codefresh: you don't need `StringifyArray`, just do: `fmt.Sprintf("some message: %v", array)`

roi-codefresh: also make sure your error messages are all lowercase

rotem-codefresh: with %v it looks ugly :(

roi-codefresh (approved):

PR #240, opened 2022-02-03
CR-9099 editable ingress host

rotem-codefresh:


noam-codefresh (approved):

cmd/commands/common.go
```diff
@@ -427,27 +444,45 @@ func getIngressHostFromCluster(ctx context.Context, opts *RuntimeInstallOptions)
...
 			} else {
-				opts.IngressHost = fmt.Sprintf("https://%s", ingress.IP)
+				foundIngressHost = fmt.Sprintf("https://%s", ingress.IP)
 			}
 			break
 		}
 	}
 
+	if store.Get().Silent {
+		log.G(ctx).Warn("Using ingress host %s", foundIngressHost)
```
roi-codefresh: Warnf**

roi-codefresh: `Warn` is without interpolation `Warnf` is with.

cmd/commands/common.go
```diff
@@ -427,27 +444,45 @@ func getIngressHostFromCluster(ctx context.Context, opts *RuntimeInstallOptions)
...
 
+	var foundIngressHost string
+
 	for _, s := range ServicesList.Items {
-		if s.ObjectMeta.Name == opts.IngressController {
+		if s.ObjectMeta.Name == opts.IngressController && s.Spec.Type == "LoadBalancer" {
 			ingress := s.Status.LoadBalancer.Ingress[0]
 			if ingress.Hostname != "" {
-				opts.IngressHost = fmt.Sprintf("https://%s", ingress.Hostname)
+				foundIngressHost = fmt.Sprintf("https://%s", ingress.Hostname)
```
roi-codefresh, line 455: why don't you break out of here if you found a result?

PR #244, opened 2022-02-03
CR-9195 validate git source token

rotem-codefresh:


cmd/commands/common.go
```diff
@@ -282,23 +271,42 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions, verify bool
...
+			opts.GitIntegrationRegistrationOpts.Token, err = cmd.Flags().GetString("git-token")
+			if err != nil {
+				return err
+			}
+		}
 	}
 
-	opts.GitIntegrationRegistrationOpts.Token, err = cmd.Flags().GetString("git-token")
-	return err
+	err = cfgit.VerifyToken(cmd.Context(), opts.InsCloneOpts.Provider, opts.GitIntegrationRegistrationOpts.Token, cfgit.PersonalToken)
```
ghost: you can just return this call

cmd/commands/common.go
```diff
@@ -282,23 +271,42 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions, verify bool
...
+	}
+
+	if  tokenFromFlag == "" {
+		if !store.Get().Silent {
+			err = getGitPATFromUserInput(cmd, opts)
+			if err != nil {
+				return err
+			}
+		} else {
+			opts.GitIntegrationRegistrationOpts.Token, err = cmd.Flags().GetString("git-token")
```
ghost: so in silent mode, we just take the value from the git-token?
doesn't it make sense to just fail in silent mode, if it is empty? if the user wants to run in silent mode, they should be super-explicit and supply the flag value.

on the other hand - in non-silent mode, maybe the default value of the flag should be the git-token (so, in the `getGitPATFromUserInput` it should be used if the user leaves it empty).
wdyt?

ghost: oh, i just now see you did the 2nd bit already.

cmd/commands/runtime.go
```diff
@@ -301,28 +303,13 @@ func runtimeInstallCommandPreRunHandler(cmd *cobra.Command, opts *RuntimeInstall
...
-	}
-
-	if opts.GsCloneOpts.Provider == "" {
-		opts.GsCloneOpts.Provider = opts.InsCloneOpts.Provider
-	}
+	opts.GsCloneOpts.Provider = opts.InsCloneOpts.Provider
+	opts.GsCloneOpts.Auth = opts.InsCloneOpts.Auth
+	opts.GsCloneOpts.Progress = opts.InsCloneOpts.Progress
+	host, orgRepo, _, _, _, suffix, _ := aputil.ParseGitUrl(opts.InsCloneOpts.Repo)
+	opts.GsCloneOpts.Repo = host + orgRepo + "_git-source" + suffix + "/resources" + "_" + opts.RuntimeName
```
ghost: i think those calls should happen in the pre-install check section

cmd/commands/common.go
rotem-codefresh: you might be right about the silent mode, but on the other hand, they can change the PAT later from the UI, as far as i understand, so maybe it is better to not add another complication in the installation? i don't know, just a thought

cmd/commands/runtime.go
rotem-codefresh: but it's not a check

ghost: sorry, i meant in the `runtimeInstallCommandPreRunHandler`
so that once the actual `RunRuntimeInstall` func is running, all the flags and options will already have their values set.

cmd/commands/common.go
ghost: ok

cmd/commands/runtime.go
ghost: my bad. i should have looked at it more closely. sorry

noam-codefresh (approved):

cmd/commands/common.go
```diff
@@ -105,14 +105,13 @@ func askUserIfToInstallDemoResources(cmd *cobra.Command, sampleInstall *bool) er
 
 		_, result, err := prompt.Run()
 		if err != nil {
-			return fmt.Errorf("Prompt error: %w", err)
+			log.G().Fatalf("prompt error: %s", err.Error())
```
roi-codefresh: no `Fatalf` is allowed unless there is a really good reason

cmd/commands/common.go
```diff
@@ -146,16 +142,15 @@ func ensureRepo(cmd *cobra.Command, runtimeName string, cloneOpts *git.CloneOpti
...
 
-func getRepoFromUserInput(cmd *cobra.Command) error {
+func getRepoFromUserInput(cmd *cobra.Command) {
 	repoPrompt := promptui.Prompt{
 		Label: "Repository URL",
 	}
 	repoInput, err := repoPrompt.Run()
 	if err != nil {
-		return fmt.Errorf("Prompt error: %w", err)
+		log.G().Fatalf("prompt error: %s", err.Error())
```
roi-codefresh: why all these `Fatalf` all of a sudden?

pkg/git/git.go
```diff
@@ -23,14 +23,27 @@ import (
 	"github.com/codefresh-io/cli-v2/pkg/log"
 )
 
+type TokenTypes string
```
roi-codefresh, line 26: Generally I think it's better design to leave the token types out of this package and instead just send in the required scopes from outside. A `git` package should not care about `runtime` vs `personal` token.
But you can leave it.

cmd/commands/runtime.go
```diff
@@ -307,22 +312,11 @@ func runtimeInstallCommandPreRunHandler(cmd *cobra.Command, opts *RuntimeInstall
...
```
roi-codefresh: Maybe move this to a function that explains what is done here, something like `initializeGitSourceCloneOpts(opts)`

cmd/commands/runtime.go
```diff
@@ -187,7 +187,10 @@ func NewRuntimeInstallCommand() *cobra.Command {
 			err := runtimeInstallCommandPreRunHandler(cmd, &installationOpts)
 			handleCliStep(reporter.InstallPhasePreCheckFinish, "Finished pre installation checks", err, false)
 			if err != nil {
-				return util.DecorateErrorWithDocsLink(fmt.Errorf("Pre installation error: %w", err), store.Get().RequirementsLink)
+				if err.Error() != "^C" {
```
roi-codefresh: Is there a better way to recognize this? pretty sure it would be different between different operating system. Does the error have a type?

cmd/commands/common.go
```diff
@@ -282,23 +288,37 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions, verify bool
 
 func ensureGitPAT(cmd *cobra.Command, opts *RuntimeInstallOptions) error {
 	var err error
-	if !store.Get().Silent {
-		return getGitPATFromUserInput(cmd, opts)
+	tokenFromFlag, err := cmd.Flags().GetString("personal-git-token")
```
roi-codefresh: This should be taken from the `opts *RuntimeInstallOptions` like the rest of the `ensure..()` functions. If tomorrow we change the flag's name it would be a pain.

cmd/commands/common.go
```diff
@@ -282,23 +288,37 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions, verify bool
...
-	opts.GitIntegrationRegistrationOpts.Token, err = cmd.Flags().GetString("git-token")
-	return err
+	if  tokenFromFlag == "" {
+		if !store.Get().Silent {
+			err = getGitPATFromUserInput(cmd, opts)
+			if err != nil {
+				return err
+			}
+		} else {
+			opts.GitIntegrationRegistrationOpts.Token, err = cmd.Flags().GetString("git-token")
```
roi-codefresh: same here, don't get the value like that, take it from `opts`

roi-codefresh: Also, I kind of agree with noam here, this might be confusing, so lets add an info log here `Using runtime token as personal user token` or something like that.

roi-codefresh (approved):

PR #246, opened 2022-02-07
Cr 7868

victor-khazanov:


pkg/runtime/runtime.go
```diff
@@ -80,8 +80,10 @@ func Download(version *semver.Version, name string) (*Runtime, error) {
 	)
 
 	devMode := false
-	if strings.HasPrefix(store.RuntimeDefURL, "http") {
-		urlString := store.RuntimeDefURL
+	//if strings.HasPrefix(store.RuntimeDefURL, "http") {
+	if strings.HasPrefix("https://github.com/codefresh-io/cli-v2/blob/main/manifests/runtime.yaml", "http") {
```
daniel-codefresh: we need to remember to change this back

cmd/commands/git-source.go
```diff
@@ -663,6 +807,10 @@ func createGithubExamplePipeline(opts *gitSourceGithubExampleOptions) error {
 	if err != nil {
 		return fmt.Errorf("failed to write yaml of github example ingress. Error: %w", err)
 	}
+	err = cleanUpRedundantFields(ingressFilePath, opts.gsCloneOpts.FS, ingress)
+	if err != nil {
+		return fmt.Errorf("failed to clean up redundant fields of githup ingress YAML. Error: %w", err)
```
daniel-codefresh: maybe we can either ignore these errors (like you did in line 798) or just log them instead of returning an error and stopping the whole flow - worst case the user will have some redundant fields in his yaml like we have today, WDYT?

cmd/commands/git-source.go
```diff
@@ -273,10 +275,15 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
 	sensorFilePath := opts.gsFs.Join(opts.gsCloneOpts.Path(), store.Get().CronExampleSensorFileName)
```
daniel-codefresh, line 273: I know its not related to your task, but any chance you could change the names of the demo yaml files in the store? I want to improve consistency in the demo resources naming conventions because right now it looks weird:
`s.CronExampleWfTemplateFileName = "workflow-template.hello-world.yaml" ---> "hello-world.workflow-template.yaml"`
`s.CronExampleEventSourceFileName = "event-source.calendar.yaml" ---> "cron.event-source.yaml"`
`s.CronExampleSensorFileName = "sensor.cron.yaml" ---> "cron.sensor.yaml"`

Also if I'm not wrong the following variable is not used anywhere and can be removed completely from the store:
`GithubExampleWfTemplateFileName`

cmd/commands/git-source.go
```diff
@@ -294,7 +301,139 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
...
+
+func cleanUpRedundantFields(filename string, gsFs fs.FS, o ...interface{}) error {
+
+	type obj = map[string]interface{}
+
+	data, err := gsFs.ReadFile(filename)
+	if err != nil {
+		return fmt.Errorf("failed to read  cron example eventSorce yaml file. Error: %w", err)
+	}
+	crd := make(map[string]interface{})
```
daniel-codefresh: you can use the `obj` type which you defined in line 314 (and maybe it would be better to make it global so you can reuse it in the rest of the functions you've created as well)

cmd/commands/git-source.go
```diff
@@ -294,7 +301,139 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
...
+		return fmt.Errorf("failed to marshal cleaned up eventSorce yaml file. Error: %w", err)
+	}
+	err = billyUtils.WriteFile(gsFs, filename, data, 0666)
+	if err != nil {
+		return fmt.Errorf("failed to write cleaned up eventSorce yaml file. Error: %w", err)
+	}
 	return nil
+
+}
+func deleteIfExistsFromMap(m map[string]interface{}, key string) {
```
daniel-codefresh: if I'm not wrong `delete()` is safe to use even if a certain key does not exist so I think we can use it directly

cmd/commands/git-source.go
```diff
@@ -294,7 +301,139 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
...
+
+	deleteIfExistsFromMap(crd, "status")
+	metadata := crd["metadata"].(obj)
+	deleteIfExistsFromMap(metadata, "creationTimestamp")
+
+	data, err = yaml.Marshal(crd)
+	if err != nil {
+		return fmt.Errorf("failed to marshal cleaned up eventSorce yaml file. Error: %w", err)
+	}
+	err = billyUtils.WriteFile(gsFs, filename, data, 0666)
```
daniel-codefresh: wouldn't gsFs.WriteYamls() work here as well?

daniel-codefresh: BTW, any chance that a clean up would work before writing to the yaml for the first time? 
I think that if we manage to convert the original struct (for example lets say a `Calendar EventSource` struct) to a `map[string]interface{}`, so we can do the whole clean up before writing each of the crds for the first time and then we can avoid redundant reads and writes for each clean up, WDYT?

cmd/commands/git-source.go
```diff
@@ -294,7 +301,139 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
...
+
+}
+func deleteIfExistsFromMap(m map[string]interface{}, key string) {
+	_, ok := m[key]
+	if ok {
+		delete(m, key)
+	}
+}
+
+func nestedMapLookup(m map[string]interface{}, ks ...string) (rval interface{}, mm map[string]interface{}) {
```
daniel-codefresh: Nice! In general this method could be very useful for other cases as well IMO, maybe it would be better to extract it to a common place as a common utility?

daniel-codefresh: Also I would suggest to add some test cases for this method, just to make sure

cmd/commands/git-source.go
```diff
@@ -294,7 +301,139 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
...
+	var ok bool
+
+	if len(ks) == 0 {
+		return nil, nil
+	}
+	if rval, ok = m[ks[0]]; !ok {
+		return nil, nil
+	} else if len(ks) == 1 {
+		return rval, m
```
daniel-codefresh: Hmm not sure I fully understand the purpose of returning both `rval` and `m` ? shouldn't this method just return a single value which is a pointer to the requested target field (something like `lodash get()` ) ? or do we need both `rval` and `m` ?

cmd/commands/git-source.go
```diff
@@ -294,7 +301,139 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
...
 
+	err = cleanUpRedundantFields(sensorFilePath, opts.gsCloneOpts.FS, sensor)
+	if err != nil {
+		return fmt.Errorf("failed to clean up redundant fields of eventSource. Error: %w", err)
+	}
+
+	return nil
+}
+
+func cleanUpRedundantFields(filename string, gsFs fs.FS, o ...interface{}) error {
```
daniel-codefresh: In general the logic LGTM but it seems like this function tries to remove all possible redundant fields for all types of resources which seems a bit complicated and can grow in size overtime. Also, this way, we are performing lots of unnecessary operations when for example we are sending here a cron eventsource and trying to remove fields which will only be present on a github eventsource. Maybe we could split this function into multiple functions like `cleanUpCronEventsource`, `cleanUpGithubEventsource`, etc, and have an additional function that can be reused for deleting common redundant fields? WDYT?

daniel-codefresh (changes requested):

daniel-codefresh (approved):

cmd/commands/git-source.go
```diff
@@ -273,7 +273,15 @@ func createCronExamplePipeline(opts *gitSourceCronExampleOptions) error {
 	sensorFilePath := opts.gsFs.Join(opts.gsCloneOpts.Path(), store.Get().CronExampleSensorFileName)
 
 	eventSource := createCronExampleEventSource()
-	err = opts.gsCloneOpts.FS.WriteYamls(eventSourceFilePath, eventSource)
+	eventSourceRedundanded, err := cleanUpRedundantFieldsFromCalendarEventSource(&eventSource)
```
roi-codefresh: I cannot understand why we are doing this after the `eventSource` was declared, and then we need to `marshal` and `unmarshal` multiple times. Why not just mutate it as an object inside the `createCronExampleEventSource` function?

roi-codefresh (approved):

PR #263, opened 2022-02-20
CR-9612 support nginx enterprise version controller

rotem-codefresh:


cmd/commands/runtime.go
```diff
@@ -440,49 +451,63 @@ func ensureIngressClass(ctx context.Context, opts *RuntimeInstallOptions) error
...
-		return nil
-	}
+	return nil
+}
 
-	return fmt.Errorf("please add the --ingress-class flag and define its value")
+func getIngressControllerName(controllerType ingressControllerType, className string) string {
+	switch controllerType {
+	case NginxCommunity:
+		return fmt.Sprintf("%s-controller", string(NginxCommunity))
```
ghost: so isn't that just `k8s.io/ingress-nginx-controller`?

manifests/runtime.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 spec:
   defVersion: 1.0.1
   version: 0.0.239
-  bootstrapSpecifier: github.com/codefresh-io/cli-v2/manifests/argo-cd
+  bootstrapSpecifier: ./manifests/argo-cd
```
ghost: make sure not to merge that

cmd/commands/git-source.go
```diff
@@ -92,6 +93,7 @@ type (
 		hostName     string
 		ingressHost  string
 		ingressClass string
+		minion       bool
```
ghost: i think the field name should be more generic, like `nginxEnt` or something. at the end, when creating an Ingress resource with that field set to `true`, it results in the minion label (and/or some other stuff).

manifests/argo-cd/resource.customizations.health.networking.k8s.io_Ingress
```diff
@@ -0,0 +1,17 @@
+hs = {}
+if obj.metadata.annotations["nginx.org/mergeable-ingress-type"] == "minion" then
+  hs.status = "Healthy"
+  return hs
+end
+if obj.metadata.annotations["nginx.org/mergeable-ingress-type"] == "master" then
+  hs.status = "Healthy"
+  return hs
+end
```
ghost: is there a way in lua to just check if `obj.metadata.annotations["nginx.org/mergeable-ingress-type"]` exists?
anyway - really impressive lua stuff. well done.

cmd/commands/git-source.go
rotem-codefresh: i thought that maybe there will be other controllers that will also use this method, so i wanted to keep it not associated to a specific controller type

cmd/commands/runtime.go
rotem-codefresh: yes, `ingress-nginx-controller` you think it better to just write it as it is?

manifests/argo-cd/resource.customizations.health.networking.k8s.io_Ingress
rotem-codefresh: i can try to figure out it there is, do you mean so that we can just check if it exists and that will be the condition?

cmd/commands/runtime.go
ghost: i think so. why have a type which is basically a string, and then cast it to string, and into a format-string?

manifests/argo-cd/resource.customizations.health.networking.k8s.io_Ingress
ghost: yeah, instead of checking if its "minion", and later check if it's "master", and basically do the same in both conditions.

noam-codefresh (approved):

cmd/commands/runtime.go
```diff
@@ -134,12 +135,22 @@ type (
...
+		Type ingressControllerType
+	}
 )
 
 const (
 	Success summaryLogLevels = "Success"
 	Failed  summaryLogLevels = "Failed"
 	Info    summaryLogLevels = "Info"
+
+	NginxCommunity  ingressControllerType = "k8s.io/ingress-nginx"
```
roi-codefresh: Prefix the name with the type so you can distinguish between them in global scope.
`IngressControllerNginxCommunity`
`IngressControllerNginxEnterprise`

cmd/commands/runtime.go
```diff
@@ -458,49 +469,69 @@ func ensureIngressClass(ctx context.Context, opts *RuntimeInstallOptions) error
...
+	opts.IngressController = ingressClassNameToController[opts.IngressClass].Name
+	opts.IngressControllerType = ingressClassNameToController[opts.IngressClass].Type
+
+	return nil
+}
+
+func getIngressControllerName(controllerType ingressControllerType, className string) string {
+	switch controllerType {
+	case NginxCommunity:
+		return "ingress-nginx-controller"
```
roi-codefresh, line 527: this name doesn't sound right. I'm pretty sure it should be `nginx-ingress-controller`

cmd/commands/runtime.go
```diff
@@ -1797,7 +1872,16 @@ func configureAppProxy(ctx context.Context, opts *RuntimeInstallOptions, rt *run
 					ServicePort: store.Get().AppProxyServicePort,
 				},
 			},
-		})
+		}
+
+		if opts.IngressControllerType == NginxEnterprise {
```
roi-codefresh: I don't want those lines to repeat >3 times, need to move to a function.

roi-codefresh (approved):

cmd/commands/runtime.go
rotem-codefresh: no, this is the right name :)

cmd/commands/runtime.go
rotem-codefresh: they are not repeating. the logic is different in the 3 cases

PR #271, opened 2022-02-21
Added shell completion subcommand

christianh814:
Adding shell completion subcommand to follow other cobra based tools.

Signed-off-by: Christian Hernandez <chernand@redhat.com>


roi-codefresh:
Thank you very much @christianh814  ! 🥇

PR #282, opened 2022-02-24
CR-9610

elad-codefresh:


roi-codefresh (approved):

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 )
 
 replace (
+	github.com/argoproj-labs/argocd-autopilot => /Users/elad-codefresh/work/argocd-autopilot
```
elad-codefresh: will remove

pkg/store/store.go
```diff
@@ -219,6 +222,7 @@ func init() {
 	s.NetworkTesterImage = "quay.io/codefresh/cf-venona-network-tester:latest"
 	s.MinKubeVersion = "v1.18.0"
 	s.MaxKubeVersion = "v1.21.9"
+	s.CFInternalGitSources = []string{"marketplace-git-source"}
```
elad-codefresh: gs logic

pkg/store/store.go
```diff
@@ -131,6 +132,7 @@ type Store struct {
 	NetworkTesterImage                   string
 	MinKubeVersion                       string
 	MaxKubeVersion                       string
+	CFInternalGitSources                 []string
```
elad-codefresh, line 135: gs logic

cmd/commands/git-source.go
```diff
@@ -226,6 +226,9 @@ func RunGitSourceCreate(ctx context.Context, opts *GitSourceCreateOptions) error
 		URL:  opts.GsCloneOpts.Repo,
 	}
 
+	_, isInternal := FindIndexIfExists(store.Get().CFInternalGitSources, appDef.Name)
```
elad-codefresh: gs logic

cmd/commands/git-source.go
```diff
@@ -19,7 +19,7 @@ import (
 	"os"
 	"strings"
 	"time"
-
+	"strconv"
```
elad-codefresh: gs logic

cmd/commands/common.go
```diff
@@ -585,3 +585,13 @@ func askUserIfToProceedWithInsecure(ctx context.Context) error {
 	}
 	return nil
 }
+
+func FindIndexIfExists(slice []string, val string) (int, bool) {
```
elad-codefresh: gs logic - and for future use

roi-codefresh: move to pkg/util/util.go rename to `StringIndexOf`, no need to return `bool` it is a standard that those functions that return index return `-1` if not found.

cmd/commands/git-source.go
```diff
@@ -226,6 +226,9 @@ func RunGitSourceCreate(ctx context.Context, opts *GitSourceCreateOptions) error
 		URL:  opts.GsCloneOpts.Repo,
 	}
 
+	_, isInternal := FindIndexIfExists(store.Get().CFInternalGitSources, appDef.Name)  
+	appDef.IsInternal = strconv.FormatBool(isInternal)
```
roi-codefresh: after you fix the util function you can do:
```
appDef.IsInternal = util.StringIndexOf(store.Get().CFInternalGitSources, appDef.Name) > -1
```

pkg/store/store.go
roi-codefresh: u can use `s.MarketplaceGitSourceName` as value here :)

roi-codefresh (approved):

cmd/commands/git-source.go
```diff
@@ -444,13 +447,16 @@ func NewGitSourceListCommand() *cobra.Command {
 				return fmt.Errorf("must enter runtime name")
 			}
 
-			return RunGitSourceList(cmd.Context(), args[0])
+			return RunGitSourceList(cmd.Context(), args[0], cmd)
 		},
 	}
+
+	cmd.Flags().BoolVar(&includeInternal, "include-internal", false, "If true, will include the Codefresh-internal git-sources")
```
roi-codefresh: very good.

roi-codefresh:
```suggestion
	cmd.Flags().BoolVar(&includeInternal, "include-internal", false, "If true, will include the Codefresh internal git-sources")
```

cmd/commands/git-source.go
```diff
@@ -444,13 +447,16 @@ func NewGitSourceListCommand() *cobra.Command {
...
+
 	return cmd
 }
 
-func RunGitSourceList(ctx context.Context, runtimeName string) error {
+func RunGitSourceList(ctx context.Context, runtimeName string, cmd *cobra.Command) error {
```
roi-codefresh: why not just pass in the `bool`? I am super confused by this :/

cmd/commands/git-source.go
```diff
@@ -474,6 +480,11 @@ func RunGitSourceList(ctx context.Context, runtimeName string) error {
 
 	for _, gs := range gitSources {
 		name := gs.Metadata.Name
+		nameWithoutRuntimePrefix := strings.Replace(name, fmt.Sprintf("%s-", runtimeName), "", 1)
+		if util.StringIndexOf(store.Get().CFInternalGitSources, nameWithoutRuntimePrefix) > -1 && !cmd.Flags().Changed("include-internal") {
```
roi-codefresh: maybe just have a util function` isInternalGitSource(gs *GitSource..) bool`

cmd/commands/runtime.go
```diff
@@ -868,7 +877,8 @@ func installComponents(ctx context.Context, opts *RuntimeInstallOptions, rt *run
 				version:      "v1alpha1",
 			},
 		},
-		saName: store.Get().RolloutReporterServiceAccount,
+		saName:     store.Get().RolloutReporterServiceAccount,
+		IsInternal: util.StringIndexOf(store.Get().CFInternalReporters, store.Get().RolloutReporterName) > -1,
```
roi-codefresh: I don't understand this. it worked without it, why do we need this?

elad-codefresh: I don't think it worked without it, it was one of the things I added, in order to include the label in the createReporter func

cmd/commands/git-source.go
elad-codefresh: won't this (StringIndexOf) be a useful function for possible future uses?

cmd/commands/git-source.go
elad-codefresh: I can pass the bool instead, and do this operation one level sooner, np

cmd/commands/git-source.go
```diff
@@ -474,6 +480,11 @@ func RunGitSourceList(ctx context.Context, runtimeName string) error {
 
 	for _, gs := range gitSources {
 		name := gs.Metadata.Name
+		nameWithoutRuntimePrefix := strings.Replace(name, fmt.Sprintf("%s-", runtimeName), "", 1)
```
roi-codefresh: strings.StripPrefix is a better fit

cmd/commands/runtime.go
roi-codefresh: what I don't understand is - which reporter is not internal? I thin this should be assumed that all of them are and just set it to true inside `createReporter` func

cmd/commands/git-source.go
elad-codefresh: I see that StripPrefix is part of net/http and not strings lib, am I missing something?

cmd/commands/runtime.go
```diff
@@ -1651,9 +1662,10 @@ func createEventsReporter(ctx context.Context, cloneOpts *git.CloneOptions, opts
 
 	resPath := cloneOpts.FS.Join(apstore.Default.AppsDir, store.Get().EventsReporterName, opts.RuntimeName, "resources")
 	appDef := &runtime.AppDef{
-		Name: store.Get().EventsReporterName,
-		Type: application.AppTypeDirectory,
-		URL:  cloneOpts.URL() + "/" + resPath,
+		Name:       store.Get().EventsReporterName,
+		Type:       application.AppTypeDirectory,
+		URL:        cloneOpts.URL() + "/" + resPath,
+		IsInternal: util.StringIndexOf(store.Get().CFInternalReporters, store.Get().EventsReporterName) > -1,
```
roi-codefresh: just `true` here.

roi-codefresh (approved):

PR #284, opened 2022-02-27
CR-9722

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -1306,6 +1328,15 @@ func RunRuntimeUninstall(ctx context.Context, opts *RuntimeUninstallOptions) err
 
 	log.G(ctx).Infof("Uninstalling runtime \"%s\" - this process may take a few minutes...", opts.RuntimeName)
 
+	err = removeGitIntegrations(ctx, opts)
+	handleCliStep(reporter.UninstallStepRemoveGitIntegrations, "Removing git integrations", err, true)
+	if err != nil {
+		if !opts.Force {
+			summaryArr = append(summaryArr, summaryLog{"you can attempt to uninstall again with the \"--force\" flag", Info})
```
roi-codefresh: Actually I think it would be better if it did affect and actually skipped this `removeGitIntegrations` step.

pkg/reporter/reporter.go
```diff
@@ -104,6 +104,7 @@ const (
 	UninstallPhaseStart                        CliStep = "uninstall.run.phase.start"
 	UninstallStepCheckRuntimeExists            CliStep = "uninstall.run.step.check-runtime-exists"
 	UninstallStepUninstallRepo                 CliStep = "uninstall.run.step.uninstall-repo"
+	UninstallStepRemoveGitIntegrations         CliStep = "uninstall.run.step.remove-git-integrations"
```
roi-codefresh, line 107: @danielm-codefresh please review the addition of a new step to the uninstall flow.

cmd/commands/runtime.go
```diff
@@ -781,6 +781,28 @@ func createGitIntegration(ctx context.Context, opts *RuntimeInstallOptions) erro
...
+	integrations, err := appProxyClient.GitIntegrations().List(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get list of git integrations: %w", err)
+	}
+
+	for _, intg := range integrations {
+		if err = RunGitIntegrationRemoveCommand(ctx, appProxyClient, intg.Name); err != nil {
+			command := util.Doc(fmt.Sprintf("\t<BIN> integration git remove %s", intg.Name))
+	
+			return fmt.Errorf(`%w. You can try to create it manually by running: %s`, err, command)
```
roi-codefresh: The message here is wrong. and actually I think you can just straight up return the error from `RunGitIntegrationRemoveCommand` because it's probably descriptive enough...

roi-codefresh (approved):

PR #286, opened 2022-02-28
CR-9389

victor-khazanov:


manifests/app-proxy/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ kind: Kustomization
 images:
   - name: quay.io/codefresh/cap-app-proxy
     newName: quay.io/codefresh/cap-app-proxy
-    newTag: v0.0.28
+    newTag: v0.0.29
```
danielm-codefresh, line 6: I see that this tag doesn't exist in quay yet, please make sure to push it before bumping

roi-codefresh (approved):

PR #288, opened 2022-02-28
implement cluster add and list

ghost:
still requires the current kube context to be set correctly before executing the command
might end up adding clusters in a different way completely

### DIFF
--- a/build/add-cluster-script.yaml
+++ b/build/add-cluster-script.yaml
@@ -1,0 +1,75 @@
+version: '1.0'
+mode: parallel
+
+stages:
+  - Prepare
+  - Test
+  - Release
+
+steps:
+  main_clone:
+    stage: Prepare
+    title: clone repository
+    type: git-clone
+    git: cf_github
+    repo: ${{CF_REPO_OWNER}}/${{CF_REPO_NAME}}
+    revision: ${{CF_BRANCH}}
+
+  prepare_env_vars:
+    stage: Prepare
+    title: prepare-env
+    image: codefresh/semver:latest
+    working_directory: manifests/add-cluster
+    commands:
+      - cf_export CHART_VERSION=$(yq ".version" helm/Chart.yaml | tr -d '"')
+      - cf_export IMAGE_VERSION=$(yq ".appVersion" helm/Chart.yaml | tr -d '"')
+    when:
+      steps:
+        - name: main_clone
+          on:
+            - success
+  
+  compare_versions:
+    stage: Test
+    title: Compare versions
+    image: codefresh/semver:latest
+    commands:
+      - hack/compare-versions.sh
+    when:
+      steps:
+        - name: prepare_env_vars
+          on:
+            - success
+
+  build_image:
+    stage: Release
+    type: build
+    title: build and push add-cluster image
+    working_directory: manifests/add-cluster
+    image_name: codefresh/csdp-add-cluster
+    tag: ${{IMAGE_VERSION}}
+    tags:
+      - latest
+      - ${{IMAGE_VERSION}}
+    registry: ${{REGISTRY_INTEGRATION_QUAY}}
+    buildkit: true
+    when:
+      steps:
+        - name: compare_versions
+          on:
+            - success
+
+  publish_chart:
+    stage: Release
+    title: publish helm chart
+    image: quay.io/codefresh/golang-ci-helper:latest # change it
+    working_directory: manifests/add-cluster
+    environment:
+      - CHART_VERSION=${{CHART_VERSION}}
+    commands:
+      - echo "inside publish helm chart ${CHART_VERSION}"
+    when:
+      steps:
+        - name: build_image
+          on:
+            - success
--- a/cmd/commands/cluster.go
+++ b/cmd/commands/cluster.go
@@ -1,0 +1,257 @@
+// Copyright 2022 The Codefresh Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/codefresh-io/cli-v2/pkg/log"
+	"github.com/codefresh-io/cli-v2/pkg/store"
+	"github.com/codefresh-io/cli-v2/pkg/util"
+	cdutil "github.com/codefresh-io/cli-v2/pkg/util/cd"
+	kustutil "github.com/codefresh-io/cli-v2/pkg/util/kust"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/argoproj-labs/argocd-autopilot/pkg/kube"
+	"github.com/juju/ansiterm"
+	"github.com/spf13/cobra"
+	kusttypes "sigs.k8s.io/kustomize/api/types"
+)
+
+type (
+	ClusterAddOptions struct {
+		runtimeName string
+		kubeContext string
+		dryRun      bool
+		kubeFactory kube.Factory
+	}
+)
+
+var minAddClusterSupportedVersion = semver.MustParse("0.0.283")
+
+func NewClusterCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Hidden:            true, // until app-proxy is working correctly
+		Use:               "cluster",
+		Short:             "Manage clusters of Codefresh runtimes",
+		PersistentPreRunE: cfConfig.RequireAuthentication,
+		Args:              cobra.NoArgs, // Workaround for subcommand usage errors. See: https://github.com/spf13/cobra/issues/706
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.HelpFunc()(cmd, args)
+			exit(1)
+		},
+	}
+
+	cmd.AddCommand(NewClusterAddCommand())
+	cmd.AddCommand(NewClusterListCommand())
+
+	return cmd
+}
+
+func NewClusterAddCommand() *cobra.Command {
+	var (
+		opts ClusterAddOptions
+		err  error
+	)
+
+	cmd := &cobra.Command{
+		Use:     "add RUNTIME_NAME",
+		Short:   "Add a cluster to a given runtime",
+		Args:    cobra.MaximumNArgs(1),
+		Example: util.Doc(`<BIN> cluster add my-runtime --context my-context`),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+
+			ctx := cmd.Context()
+
+			opts.runtimeName, err = ensureRuntimeName(ctx, args)
+			if err != nil {
+				return err
+			}
+
+			opts.kubeContext, err = ensureKubeContextName(cmd.Flag("context"))
+			return err
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runClusterAdd(cmd.Context(), &opts)
+		},
+	}
+
+	cmd.Flags().BoolVar(&opts.dryRun, "dry-run", false, "")
+	opts.kubeFactory = kube.AddFlags(cmd.Flags())
+	die(err)
+
+	return cmd
+}
+
+func runClusterAdd(ctx context.Context, opts *ClusterAddOptions) error {
+	runtime, err := cfConfig.NewClient().V2().Runtime().Get(ctx, opts.runtimeName)
+	if err != nil {
+		return err
+	}
+
+	if runtime.RuntimeVersion == nil {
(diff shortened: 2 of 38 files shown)


build/add-cluster-script.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+  publish_chart:
+    stage: Release
+    title: publish helm chart
+    image: quay.io/codefresh/golang-ci-helper:latest # change it
+    working_directory: manifests/add-cluster
+    environment:
+      - CHART_VERSION=${{CHART_VERSION}}
+    commands:
+      - echo "inside publish helm chart ${CHART_VERSION}"
```
roi-codefresh, line 70: so nothing here yet?

cmd/commands/cluster.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func NewClusterAddCommand() *cobra.Command {
+	var (
+		opts ClusterAddOptions
+		err  error
+	)
+
+	cmd := &cobra.Command{
+		Use:     "add RUNTIME_NAME --context CONTEXT_NAME",
```
roi-codefresh: should not have flags in the `USE` string, only positional params.
This should be an example.

cmd/commands/cluster.go
```diff
@@ -0,0 +1,258 @@
...
+		Short:   "Add a cluster to a given runtime",
+		Args:    cobra.MaximumNArgs(2),
```
roi-codefresh: Why 2 args? I see only one positional arg.

build/add-cluster-script.yaml
ghost: yeah, once Itai helps me figure out where and how to publish, i will add it. but until then - the kustomize + cli way can already work

cmd/commands/cluster.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func runClusterAdd(ctx context.Context, opts *ClusterAddOptions) error {
+	runtime, err := cfConfig.NewClient().V2().Runtime().Get(ctx, opts.runtimeName)
+	if err != nil {
+		return err
+	}
+
+	if runtime.RuntimeVersion == nil {
+		return fmt.Errorf("runtme \"%s\" has no version", opts.runtimeName)
```
roi-codefresh:
```suggestion
		return fmt.Errorf("runtime \"%s\" has no version", opts.runtimeName)
```

cmd/commands/cluster.go
```diff
@@ -0,0 +1,258 @@
...
+type (
+	ClusterAddOptions struct {
+		runtimeName string
+		kubeContext string
+		dryRun      bool
+		kubeFactory kube.Factory
+	}
+)
+
+var minimalSupportedVersion = semver.MustParse("99.99.99") // to be updated after app-proxy is ready
```
roi-codefresh: this is package scope, so not the best name

cmd/commands/cluster.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if opts.dryRun {
+		fmt.Println(string(manifests))
+		return nil
+	}
+
+	return opts.kubeFactory.Apply(ctx, "", manifests)
+}
+
+func createAddClusterKustomization(ingressUrl, contextName, server, csdpToken string, version *string) *kusttypes.Kustomization {
```
roi-codefresh:
```suggestion
func createAddClusterKustomization(ingressUrl, contextName, server, csdpToken string, version string) *kusttypes.Kustomization {
```

Think it's better to dereference in the calling function

cmd/commands/cluster.go
```diff
@@ -0,0 +1,258 @@
...
+	resourceUrl := store.AddClusterDefURL
+	if strings.HasPrefix(resourceUrl, "http") && version != nil {
+		resourceUrl = fmt.Sprintf("%s?def=%s", resourceUrl, *version)
```
roi-codefresh:
```suggestion
		resourceUrl = fmt.Sprintf("%s?ref=%s", resourceUrl, *version)
```

Is it supposed to be `ref` here?

ghost: mos def

cmd/commands/cluster.go
ghost: do you think i should just rename it? or also move it into some other more specific scope? maybe inside `runClusterAdd`?

hack/release.sh
```diff
@@ -17,24 +17,18 @@ fi
 if [[ -z "$PRERELEASE" ]]; then
     PRERELEASE=false
 fi
-    
+
 echo "uploading files:"
-ls -1a ./dist/*.tar.gz ./dist/*.sha256
+ls -1a ./dist/*.tar.gz ./dist/*.sha256 ./manifests/runtime.yaml
```
roi-codefresh, line 22: why do we suddenly need this?

manifests/add-cluster/add-cluster.sh
```diff
@@ -0,0 +1,48 @@
...
+KUBE_CONFIG_DATA=$(kubectl config view --minify --flatten --context=${CONTEXT_NAME} | base64)
+
+# get ServiceAccount token
+SECRET_NAME=$(kubectl get ServiceAccount ${SERVICE_ACCOUNT_NAME} -n ${NAMESPACE} -o jsonpath='{.secrets[0].name}')
+echo "Found ServiceAccount secret ${SECRET_NAME}"
+BEARER_TOKEN=$(kubectl get secret ${SECRET_NAME} -n ${NAMESPACE} -o jsonpath='{.data.token}')
+
+
+RESPONSE=$(curl -X POST ${INGRESS_URL}/app-proxy/api/clusters \
+  -H 'Content-Type: application/json' \
```
roi-codefresh, line 35: maybe you should configure here some sort of retry. pretty sure you can do that with a flag or two.

hack/release.sh
ghost: yeah, for a while i thought i'd add a built kustomization manifest to the artifact, so i looked at the code. i just noticed that the ls command is not identical to the actual files being uploaded as artifact. it doesn't really matter

manifests/add-cluster/helm/templates/clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Values.serviceAccount | default "argocd-manager" }}-role
```
roi-codefresh: Need to concat chart unique value so those won't collide with other cluster scope resources.

manifests/add-cluster/helm/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Values.serviceAccount | default "argocd-manager" }}-role-binding
```
roi-codefresh: here too

manifests/add-cluster/helm/templates/job.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "csdp-add-cluster.fullname" . }}-job
+  namespace: {{ .Values.systemNamespace | default "kube-system" }}
+  labels:
+    {{- include "csdp-add-cluster.labels" . | nindent 4 }}
+spec:
+  ttlSecondsAfterFinished: 600 # stick around for 10m
```
roi-codefresh, line 9: where is a retry policy?

manifests/add-cluster/helm/values.yaml
```diff
@@ -0,0 +1,18 @@
+# serviceAccountName is the name of the ServiceAccount Argo-CD will use to access the cluster
+serviceAccountName: argocd-manager
+
+# ingressUrl is the url used to access the CSDP runtime
+ingressUrl: <ingressUrl>
```
roi-codefresh, line 6: Is this the base url? or need the whole path? Maybe an example?

manifests/add-cluster/helm/values.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+# csdoToken is used to authenticate against the app-proxy
```
roi-codefresh:
```suggestion
# csdpToken is used to authenticate against the app-proxy
```

manifests/add-cluster/kustomize/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: csdp-add-cluster-cm
+  namespace: kube-system
+data:
+  ingressUrl: <ingressUrl>
+  contextName: <contextName>
+  server: <server
```
roi-codefresh:
```suggestion
  server: <server>
```

manifests/add-cluster/kustomize/job.yaml
```diff
@@ -0,0 +1,52 @@
...
+  ttlSecondsAfterFinished: 600  # stick around for 10m
+  template:
+    metadata:
+      name: csdp-add-cluster-pod
+    spec:
+      serviceAccount: argocd-manager
+      restartPolicy: Never
+      containers:
+        - name: main
+          image: quay.io/codefresh/csdp-add-cluster:0.1.0
```
roi-codefresh: put latest here and replace in the `kustomization.yaml` file

manifests/add-cluster/kustomize/secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: csdp-add-cluster-secret
+type: Opaque
+data:
+  csdpToken: <csdpToken>
```
roi-codefresh, line 7: Maybe we need to think about some way to clean those after the job is done. But not crucial for now.

pkg/util/cd/util.go
```diff
@@ -39,44 +41,55 @@ type (
...
-		return "", fmt.Errorf("failed to create argocd client: %w", err)
+		return "", fmt.Errorf("failed to generate account token: %w", err)
 	}
 
-	conn, accountIf, err := argoClient.NewAccountClient()
+	return res.Token, nil
+}
+
+func GetClusterList(ctx context.Context, kubeContext, namespace string, insecure bool) (*v1alpha1.ClusterList, error) {
+	argoClient, err := createArgoClient(kubeContext, namespace, insecure)
```
roi-codefresh, line 67: Do we plan to move this logic and use the platform client to list the clusters? Because I don't think it's a good idea to read those from argocd directly. It's different from all the rest of the entities.

manifests/add-cluster/add-cluster.sh
ghost: the job itself does a staggered retry 6 times (by default). i think that should be enough

manifests/add-cluster/helm/templates/clusterrole.yaml
ghost: so basically - adding the name the user is giving in the `helm install <name> <chart>` command?

manifests/add-cluster/helm/templates/job.yaml
ghost: the default is 6. should it be explicit? or even - a higher number?

manifests/add-cluster/helm/values.yaml
ghost: is there a way to use `cf config ... ` cli commands to display the current user's api-key? or should i direct users to just create a new api-key for this purpose? or view their `~/.cfconfig` file?

manifests/add-cluster/kustomize/job.yaml
ghost: i can do that, and have the file hard-coded 0.1.0 (and the next time somebody updates the script or image, they should remember to update both the Chart.yaml and the cluster.go file).
but this way, the v1 pipeline is easily making sure the kustomize version is identical to the helm version.

manifests/add-cluster/kustomize/secret.yaml
ghost: the script is deleting the secret after it successfully finishes.
if it fails - it doesn't

pkg/util/cd/util.go
ghost: yeah - this is a temporary placeholder. it will query the runtime instead, once the add-cluster works end-to-end

manifests/add-cluster/kustomize/job.yaml
ghost: oh. unless you mean the `manifesta/add-clustetr/kustomize/kustomization.yaml` file (not the kustomization created in-code). it does make sense to have it there. i'll move it there.

manifests/add-cluster/helm/templates/clusterrole.yaml
roi-codefresh: It's the `{{ include "csdp-add-cluster.fullname" . }}` stuff you have on a number of resources already.

manifests/add-cluster/kustomize/secret.yaml
roi-codefresh: gr8

manifests/add-cluster/helm/values.yaml
roi-codefresh: I don't think it's a normal thing a user would like a cli thing to do, to show his secrets. But we should have a command to generate a populated helm values file with all of the values.

roi-codefresh (approved):

PR #311, opened 2022-03-16
CR-9648

elad-codefresh:


cmd/commands/common.go
```diff
@@ -376,6 +376,30 @@ func ensureKubeContextName(flag *pflag.Flag) (string, error) {
 	return contextName, nil
 }
 
+func validateKubeContext(ctx context.Context, kubeContextName string, force bool) error {
+	if kubeContextName == "" {
```
elad-codefresh, line 380: will be resolved later on by ap to the currentContext, therefore no validation required here

cmd/commands/runtime.go
```diff
@@ -1405,11 +1405,22 @@ func RunRuntimeUninstall(ctx context.Context, opts *RuntimeUninstallOptions) err
 	var err error
 	if !opts.SkipChecks {
 		_, err = cfConfig.NewClient().V2().Runtime().Get(ctx, opts.RuntimeName)
-	}
-	handleCliStep(reporter.UninstallStepCheckRuntimeExists, "Checking if runtime exists", err, false, true)
-	if err != nil {
-		summaryArr = append(summaryArr, summaryLog{"you can attempt to uninstall again with the \"--skip-checks\" flag", Info})
-		return err
+		handleCliStep(reporter.UninstallStepCheckRuntimeExists, "Checking if runtime exists", err, false, true)
```
elad-codefresh, line 1408: I inserted the handleCliStep of checkRuntimeExists into the scope of the !opts.SkipChecks, in order to handle both checks inside this scope. If I understand correctly, no need to handle it in case !opts.SkipChecks.

PR #354, opened 2022-04-17
Cr 8871 - ALB support

danielm-codefresh:


cmd/commands/runtime.go
```diff
@@ -1326,6 +1335,7 @@ func RunRuntimeList(ctx context.Context) error {
 			installationStatus,
 			ingressHost,
 			ingressClass,
+			ingressController,
```
roi-codefresh: I don't think it's necessary to show in the `runtime list` command

cmd/commands/runtime.go
```diff
@@ -1891,10 +1906,17 @@ func configureAppProxy(ctx context.Context, opts *RuntimeInstallOptions, rt *run
...
 		}
 
-		if opts.IngressControllerType == IngressControllerNginxEnterprise {
+		switch opts.IngressControllerType {
+		case IngressControllerNginxEnterprise:
 			ingressOptions.Annotations = map[string]string{
 				"nginx.org/mergeable-ingress-type": "minion",
 			}
+		case IngressControllerALB:
+			ingressOptions.Annotations = map[string]string{
```
roi-codefresh: I don't really like this repetition. I would have created some `IngressController` interface. something like:
```
type IngressController interface {
    // Name returns the name of the ingress controller
    Name() string
    // Decorate adds the required controller specific annotations
    Decorate(ingress)
}
```

roi-codefresh (approved):

noam-codefresh (approved):

PR #359, opened 2022-04-18
CR-10299-2

elad-codefresh:


manifests/runtime.yaml
```diff
@@ -21,3 +21,4 @@ spec:
     - name: app-proxy
       type: kustomize
       url: github.com/codefresh-io/cli-v2/manifests/app-proxy
+      wait: true
```
elad-codefresh, line 24: will test several times the e2e, to see it consistently passes, before merging.
to validate we no longer have the race condition that we probably had before.

cmd/commands/git-source.go
```diff
@@ -205,37 +208,57 @@ func NewGitSourceCreateCommand() *cobra.Command {
...
+	version, err := getRuntimeVersion(ctx, opts.RuntimeName)
 	if err != nil {
-		return fmt.Errorf("failed to clone git-source repo: %w", err)
+		return err
 	}
 
-	if opts.CreateDemoResources {
-		if err := createDemoResources(ctx, opts, gsRepo, gsFs); err != nil {
-			return fmt.Errorf("failed to create git-source demo resources: %w", err)
+	if version.LessThan(versionOfGitSourceByAppProxyRefactor) {
```
oren-codefresh, line 216: Don't we have a breaking change mechanism?  why does it have to be hard coded ?

oren-codefresh (approved):

PR #362, opened 2022-04-19
CR-11030-5

elad-codefresh:


go.mod
```diff
@@ -248,9 +248,9 @@ require (
 )
 
 replace (
+	github.com/codefresh-io/go-sdk => /Users/elad-codefresh/work/go-sdk
```
elad-codefresh: will be replaced with new sdk

cmd/commands/git-source.go
```diff
@@ -189,53 +199,67 @@ func NewGitSourceCreateCommand() *cobra.Command {
...
+		return legacyGitSourceCreate(ctx, opts)
+	}
+
+	if version.LessThan(versionOfGitSourceByAppProxyRefactor) {
+		log.G(ctx).Infof("runtime \"%s\" is using a depracated git-source api. Versions %s and up use the app-proxy for this command", opts.RuntimeName, minAddClusterSupportedVersion)
+		return legacyGitSourceCreate(ctx, opts)
 	} else {
-		if err := createPlaceholderIfNeeded(ctx, opts, gsRepo, gsFs); err != nil {
-			return fmt.Errorf("failed to create a git-source placeholder: %w", err)
+		appProxy, err := cfConfig.NewClient().AppProxy(ctx, opts.RuntimeName, store.Get().InsecureIngressHost)
```
danielm-codefresh, line 238: Please add fallback to the `legacyGitSourceCreate` if app-proxy is not accessible

cmd/commands/git-source.go
```diff
@@ -189,53 +199,67 @@ func NewGitSourceCreateCommand() *cobra.Command {
...
+		})
 
-	appDef.IsInternal = util.StringIndexOf(store.Get().CFInternalGitSources, appDef.Name) > -1
+		if err != nil {
+			return fmt.Errorf("failed to create git-source: %w", err)
+		}
 
-	if err := appDef.CreateApp(ctx, nil, opts.InsCloneOpts, opts.RuntimeName, store.Get().CFGitSourceType, opts.Include, ""); err != nil {
-		return fmt.Errorf("failed to create git-source application. Err: %w", err)
+		log.G(ctx).Infof("Successfully created git-source: \"%s\"", opts.GsName)
```
elad-codefresh, line 262: there's an issue here.
gs creation can fail in app-proxy, but still print out this success message.
it's because current app-proxy just logs the error, doesn't return it.
we need to talk about how to address that.

ghost: wouldn't a failure in app-proxy cause the graphql request to fail?

cmd/commands/git-source.go
```diff
@@ -189,53 +199,67 @@ func NewGitSourceCreateCommand() *cobra.Command {
...
 	if err != nil {
-		return fmt.Errorf("failed to clone git-source repo: %w", err)
+		return err
 	}
 
-	if opts.CreateDemoResources {
-		if err := createDemoResources(ctx, opts, gsRepo, gsFs); err != nil {
-			return fmt.Errorf("failed to create git-source demo resources: %w", err)
-		}
+	if opts.Flow == store.Get().InstallationFlow {
```
ghost, line 230: why use a `Flow string` with two possible values? wouldn't it be simpler to use some bool flag called `ForceAutopilot` or something like that? (`ForceLocal`? )

cmd/commands/git-source.go
elad-codefresh: I noticed cases where it doesn't, for example if there's a problem with the `.exec` (e.g it can't find the ap bin), it prints the error in app-proxy, but doesn't reflect that in the cli

cmd/commands/git-source.go
elad-codefresh: I thought it might come in handy in future cases, to differentiate between the two flows like we need to do in this case. no?

noam-codefresh (approved):

PR #365, opened 2022-04-25
proxy-bump-2

elad-codefresh:


manifests/app-proxy/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ kind: Kustomization
 images:
   - name: quay.io/codefresh/cap-app-proxy
     newName: quay.io/codefresh/cap-app-proxy
-    newTag: 1.1091.0
+    newTag: 1.1107.0
```
elad-codefresh: SHOULD VERIFY AFTER PUSHING PLATFORM PR

ghost: updated to 1.1112.0

PR #366, opened 2022-04-26
proxy-bump-3

elad-codefresh:


cmd/commands/git-source.go
```diff
@@ -109,7 +109,7 @@ type (
 	}
 )
 
-var versionOfGitSourceByAppProxyRefactor = semver.MustParse("0.0.326")
+var runtimeVersionOfGitSourceByAppProxyRefactor = semver.MustParse("0.0.323")
```
ghost: this should be 327, right?

ghost: also - the name of the variable doesn't make sense to me. i think it's too long.
how about something like `runtimeGitSourceSupport`? all the cli really cares is if the app-proxy is "new" enough to support git-source mutations. maybe `appProxyGitSourceSupport`? `minGitSourceSupport`?

i don't know. it's not super important. just seems like an overly long and complicated name. especially with the `Refactor` at the end - that doesn't seem like it should interest the cli. it's either supported or not.

nvm - i'll stop rambling.

noam-codefresh (approved):

cmd/commands/git-source.go
elad-codefresh: let's settle for appProxyGitSourceSupportVersion :)

PR #384, opened 2022-05-11
CR-11574

elad-codefresh:


cmd/commands/git-source.go
```diff
@@ -524,6 +524,10 @@ func RunGitSourceList(ctx context.Context, runtimeName string, includeInternal b
 			continue
 		}
 
+		if gs.Self == nil {
```
roi-codefresh, line 527: This means you will not show it at all, I wonder if this is what we want. How does UI deal with this state?
We need to have the same behavior

elad-codefresh: currently, in this situation (no self) the ui currently won't show it as well (with the recently-added descriptive error )

roi-codefresh: Oh, but it does show an error, maybe we should also print some error

roi-codefresh (approved):

cmd/commands/git-source.go
```diff
@@ -524,6 +524,12 @@ func RunGitSourceList(ctx context.Context, runtimeName string, includeInternal b
 			continue
 		}
 
+		if strings.Contains(name, "default") {
```
roi-codefresh: This is a bug

roi-codefresh (changes requested):

roi-codefresh (approved):

PR #411, opened 2022-05-27
[BE]:Handle validation check

PhilippPlotnikov:


cmd/commands/common.go
```diff
@@ -215,21 +215,31 @@ func getRuntimeNameFromUserSelect(ctx context.Context) (string, error) {
...
-		Default: defaultValue,
-		// Validate: func (value string) error {
-		// 	if value == "" {
-		// 		return fmt.Errorf("Must supply value for \"%s\"", label)
-		// 	}
+func validateRuntimeName(runtime string) error {
+	var err error
+	isValid, err := IsValidName(runtime)
+	if err != nil {
+		err = fmt.Errorf("failed to check the validity of the runtime name: %w", err)
```
```suggestion
		return fmt.Errorf("failed to validate runtime name: %w", err)
```

cmd/commands/common.go
```diff
@@ -215,21 +215,31 @@ func getRuntimeNameFromUserSelect(ctx context.Context) (string, error) {
...
+	} else if !isValid {
+		err = fmt.Errorf("runtime name cannot have any uppercase letters, must start with a character, end with character or number, and be shorter than 63 chars")
```
```suggestion
		return fmt.Errorf("runtime name cannot have any uppercase letters, must start with a character, end with character or number, and be shorter than 63 chars")
```

cmd/commands/common.go
```diff
@@ -280,10 +290,10 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions, verify bool
 	if verify {
 		err := cfgit.VerifyToken(cmd.Context(), cloneOpts.Provider, cloneOpts.Auth.Password, cfgit.RuntimeToken)
 		if err != nil {
-			return fmt.Errorf("failed to verify git token: %w", err)
+			cloneOpts.Auth.Password = ""
```
roi-codefresh, line 292: why do we need to do this?

cmd/commands/runtime.go
```diff
@@ -342,7 +342,13 @@ func runtimeInstallCommandPreRunHandler(cmd *cobra.Command, opts *RuntimeInstall
 
 	inferProviderFromRepo(opts.InsCloneOpts)
 
-	err = ensureGitToken(cmd, opts.InsCloneOpts, true)
+	handleValidationFailsWithRepeat(func() error {
+		err = ensureGitToken(cmd, opts.InsCloneOpts, true)
+		if isValidationError(err) {
```
roi-codefresh: what happens in silent mode? I think this is a bug. This only makes sense in case git token is from user input. you should move this to be validation of user input like runtime name and ingress host

PhilippPlotnikov: Didnt understand about silent mode

cmd/commands/common.go
```diff
@@ -215,21 +215,29 @@ func getRuntimeNameFromUserSelect(ctx context.Context) (string, error) {
 }
 
 func getRuntimeNameFromUserInput() (string, error) {
-	return getValueFromUserInput("Runtime name", "codefresh")
+	runtimeName, err := getValueFromUserInput("Runtime name", "codefresh", validateRuntimeName)
+	if err != nil {
+		return runtimeName, err
+	}
+	return runtimeName, err
```
ghost: if you return `runtimeName, err` either way, you there is no need for the `if`.
you can just `return getValueFromUserInput....`

cmd/commands/common.go
```diff
@@ -215,21 +215,29 @@ func getRuntimeNameFromUserSelect(ctx context.Context) (string, error) {
...
-		Default: defaultValue,
-		// Validate: func (value string) error {
-		// 	if value == "" {
-		// 		return fmt.Errorf("Must supply value for \"%s\"", label)
-		// 	}
+func validateRuntimeName(runtime string) error {
+	isValid, err := IsValidName(runtime)
+	if err != nil {
+		return fmt.Errorf("failed to validate runtime name: %w", err)
+	} else if !isValid {
```
ghost: no need for the `else` after the `return`
you can just have a separate `if`

cmd/commands/common.go
```diff
@@ -592,6 +595,21 @@ func askUserIfToProceedWithInsecure(ctx context.Context) error {
 	return nil
 }
 
+type Callback func() error
+
+func handleValidationFailsWithRepeat(callback Callback) {
+	var err error
+	for {
+		err = callback()
+		if err == nil || err == promptui.ErrInterrupt {
```
ghost: maybe you can use `isValidationError` here

noam-codefresh (approved):

PR #424, opened 2022-06-07
CR-11767 uninstalling isc when uninstalling runtime

rotem-codefresh:


cmd/commands/runtime.go
```diff
@@ -455,6 +456,12 @@ func runtimeUninstallCommandPreRunHandler(cmd *cobra.Command, args []string, opt
 		return err
 	}
 
+	opts.IscCloneOpts.Repo, err = setIscRepo(cmd.Context(), "")
```
ghost: instead of using `setIscRepo` (which is actually calling `suggestIscRepo`), this code should just use `getIscRepo`.
this func should just request the user -> account -> iscRepo value.

noam-codefresh (approved):

PR #427, opened 2022-06-12
Cr 12025 add cluster wait

ghost:
after creating add-cluster-job, poll its status and output the logs (up to 6 attempts, should max of 10:30 minutes)

### DIFF
--- a/build/add-cluster-script.yaml
+++ b/build/add-cluster-script.yaml
@@ -58,6 +58,21 @@ steps:
         - name: compare_versions
           on:
             - success
+      branch:
+        only: ["main"]
+
+  build_dev_image:
+    stage: Release
+    type: build
+    title: build and push add-cluster dev image
+    working_directory: manifests/add-cluster
+    image_name: codefresh/csdp-add-cluster
+    tag: ${{CF_BRANCH}}
+    registry: ${{REGISTRY_INTEGRATION_QUAY}}
+    buildkit: true
+    when:
+      branch:
+        ignore: ["main"]
 
   publish_chart:
     stage: Release
--- a/build/add-cluster-script.yaml
+++ b/build/add-cluster-script.yaml
@@ -71,6 +71,10 @@ steps:
     registry: ${{REGISTRY_INTEGRATION_QUAY}}
     buildkit: true
     when:
+      steps:
+        - name: compare_versions
+          on:
+            - success
       branch:
         ignore: ["main"]
 
--- a/cmd/commands/cluster.go
+++ b/cmd/commands/cluster.go
@@ -24,6 +24,7 @@ import (
 	"github.com/codefresh-io/cli-v2/pkg/log"
 	"github.com/codefresh-io/cli-v2/pkg/store"
 	"github.com/codefresh-io/cli-v2/pkg/util"
+	kubeutil "github.com/codefresh-io/cli-v2/pkg/util/kube"
 	kustutil "github.com/codefresh-io/cli-v2/pkg/util/kust"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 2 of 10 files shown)


build/add-cluster-script.yaml
```diff
@@ -58,6 +58,25 @@ steps:
...
+    tag: ${{CF_BRANCH}}
+    registry: ${{REGISTRY_INTEGRATION_QUAY}}
+    buildkit: true
+    when:
+      steps:
+        - name: compare_versions
+          on:
+            - success
+      branch:
+        ignore: ["main"]
```
ghost, line 79: pr pushes do not trigger the pipeline. but it can be used for a manual pipeline run, just to build the matching image

manifests/add-cluster/add-cluster.sh
```diff
@@ -37,10 +37,16 @@ STATUS_CODE=$(curl -X POST ${INGRESS_URL}/app-proxy/api/clusters \
...
   -s -o response -w "%{http_code}")
-cat response
 echo "STATUS_CODE: ${STATUS_CODE}"
+cat response
+echo
+
+if [[ $STATUS_CODE == 000 ]]; then
+  echo "error sending request to runtime"
+  exit 1
+fi
```
ghost, line 47: handle cases where url is completely wrong (like, when trying to use http://localhost:3017)

cmd/commands/cluster.go
```diff
@@ -139,15 +140,20 @@ func runClusterAdd(ctx context.Context, opts *ClusterAddOptions) error {
...
 		return nil
 	}
 
-	return opts.kubeFactory.Apply(ctx, manifests)
+	err = opts.kubeFactory.Apply(ctx, manifests)
+	if err != nil {
+		return fmt.Errorf("failed applying manifests to cluster: %w", err)
+	}
+
+	return kubeutil.WaitForJob(ctx, opts.kubeFactory, "kube-system", "csdp-add-cluster-job")
```
roi-codefresh: Maybe put `"csdp-add-cluster-job"` somewhere else

manifests/add-cluster/add-cluster.sh
```diff
@@ -37,10 +37,16 @@ STATUS_CODE=$(curl -X POST ${INGRESS_URL}/app-proxy/api/clusters \
...
 
-if [ $STATUS_CODE -ge 300 ]; then
+if [[ $STATUS_CODE -ge 300 ]]; then
```
roi-codefresh, line 49: I think generally you should check for `-ge 400` because 300-399 contains a bunch of valid responses no?

pkg/util/kube/kube.go
```diff
@@ -426,8 +486,8 @@ func deleteJob(ctx context.Context, client kubernetes.Interface, job *batchv1.Jo
 }
 
 func getPodByJob(ctx context.Context, client kubernetes.Interface, job *batchv1.Job) (*v1.Pod, error) {
-	pods, err := client.CoreV1().Pods(store.Get().DefaultNamespace).List(ctx, metav1.ListOptions{
-		LabelSelector: "controller-uid=" + job.GetLabels()["controller-uid"],
+	pods, err := client.CoreV1().Pods(job.GetNamespace()).List(ctx, metav1.ListOptions{
```
roi-codefresh, line 489: Does this returns the latest pod created by this job?

roi-codefresh (approved):

pkg/util/kube/kube.go
ghost: yeah, i think every time the job creates a new pod, it uses a different controller-uid (and updates the job with it)
i'll make sure

PR #431, opened 2022-06-13
CR-11438 friendly cluster name

rotem-codefresh:


cmd/commands/cluster.go
```diff
@@ -136,7 +138,13 @@ func runClusterAdd(ctx context.Context, opts *ClusterAddOptions) error {
 	}
 
 	csdpToken := cfConfig.GetCurrentContext().Token
-	k := createAddClusterKustomization(ingressUrl, opts.kubeContext, server, csdpToken, *runtime.RuntimeVersion)
+	setClusterName(opts)
```
ghost: this should happen in the `PreRunE`
when calling the actual function, all the opts should have their final value (including the validation)

noam-codefresh (approved):

noam-codefresh (approved):

PR #432, opened 2022-06-13
CR-10413 add cluster labels/annotations

rotem-codefresh:


noam-codefresh (approved):

cmd/commands/cluster.go
```diff
@@ -123,16 +130,24 @@ func newClusterAddCommand() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runClusterAdd(cmd.Context(), &opts)
 		},
 	}
 
 	cmd.Flags().StringVar(&opts.clusterName, "name", "", "Name of the cluster. If omitted, will use the context name")
+	cmd.Flags().StringToStringVar(&opts.annotations, "annotations", nil, "Set metadata annotations (e.g. --annotation key=value)")
+	cmd.Flags().StringToStringVar(&opts.labels, "labels", nil, "Set metadata labels (e.g. --label key=value)")
 	cmd.Flags().BoolVar(&opts.dryRun, "dry-run", false, "")
+	cmd.Flags().StringVar(&opts.tag, "tag", "", "[dev only] - use a specific tag of the csdp-add-cluster image")
```
ghost, line 143: this is a hidden flag, useful for dev time - if i put in something like `--tag CR-10413-cluster-labels`, it will use that as the resource in the kustomization (`https://github.com/codefresh-io/csdp-official/add-cluster/kustomize?ref=CR-10413-cluster-labels`), and also add an `images` replacement in the kustomization, to use it as the `csdp-add-cluster` tag, instead of `0.5.0` (for example)

cmd/commands/cluster.go
```diff
@@ -123,16 +130,24 @@ func newClusterAddCommand() *cobra.Command {
 			}
 
 			err = setClusterName(cmd.Context(), &opts)
+			if err != nil {
+				return err
+			}
 
-			return err
+			return nil
```
roi-codefresh: this can be reverted I guess.

roi-codefresh (approved):

PR #435, opened 2022-06-15
CR-12287

elad-codefresh:


cmd/commands/runtime.go
```diff
@@ -2349,6 +2355,36 @@ func createReporter(ctx context.Context, cloneOpts *git.CloneOptions, opts *Runt
 	return apu.PushWithMessage(ctx, r, pushMessage)
 }
 
+func editAppSet(ctx context.Context, r git.Repository, repofs fs.FS, rtName string) error {
+	projPath := repofs.Join(apstore.Default.ProjectsDir, rtName + ".yaml")
```
elad-codefresh: named it editAppSet with open-close principle in mind

PR #443, opened 2022-06-20
CR-12478 adjust uninstall to managed

rotem-codefresh:


cmd/commands/common.go
```diff
@@ -141,7 +141,7 @@ func ensureRepo(cmd *cobra.Command, runtimeName string, cloneOpts *git.CloneOpti
 	}
 
 	if cloneOpts.Repo == "" {
-		return fmt.Errorf("must enter a valid installation repository URL")
+		return fmt.Errorf("must enter a valid installation repository URL via --repo")
```
```suggestion
		return fmt.Errorf("must enter a valid installation repository URL, using --repo")
```

cmd/commands/common.go
```diff
@@ -301,6 +310,11 @@ func ensureGitToken(cmd *cobra.Command, cloneOpts *git.CloneOptions, verify bool
 			return fmt.Errorf(errMessage)
 		}
 	}
+
+	if cloneOpts.Auth.Password == "" {
+		return fmt.Errorf("must provide a git token via --git-token")
```
```suggestion
		return fmt.Errorf("must provide a git token using --git-token")
```

cmd/commands/runtime.go
```diff
@@ -1424,13 +1447,14 @@ func RunRuntimeList(ctx context.Context) error {
 	}
 
 	tb := ansiterm.NewTabWriter(os.Stdout, 0, 0, 4, ' ', 0)
-	_, err = fmt.Fprintln(tb, "NAME\tNAMESPACE\tCLUSTER\tVERSION\tSYNC_STATUS\tHEALTH_STATUS\tHEALTH_MESSAGE\tINSTALLATION_STATUS\tINGRESS_HOST\tINGRESS_CLASS")
+	_, err = fmt.Fprintln(tb, "NAME\tTYPE\tNAMESPACE\tCLUSTER\tVERSION\tSYNC_STATUS\tHEALTH_STATUS\tHEALTH_MESSAGE\tINSTALLATION_STATUS\tINGRESS_HOST\tINGRESS_CLASS")
```
roi-codefresh: I don't think we need another column for that, we already have enough columns. I think you should do it like the runtime select-i.e. have `(hosted)`. Also people that don't even have the managed runtime feature flag and can't create managed runtimes will started to see this `Hybrid` type here and have question about it.

cmd/commands/runtime.go
```diff
@@ -1569,7 +1592,10 @@ func NewRuntimeUninstallCommand() *cobra.Command {
 	cmd.Flags().BoolVar(&opts.FastExit, "fast-exit", false, "If true, will not wait for deletion of cluster resources. This means that full resource deletion will not be verified")
 	cmd.Flags().BoolVar(&opts.DisableTelemetry, "disable-telemetry", false, "If true, will disable the analytics reporting for the uninstall process")
 
-	opts.CloneOpts = apu.AddCloneFlags(cmd, &apu.CloneFlagsOptions{CloneForWrite: true})
+	opts.CloneOpts = apu.AddCloneFlags(cmd, &apu.CloneFlagsOptions{
+		CloneForWrite: true,
+		Optional:      true,
```
roi-codefresh, line 1587: If we changed that to optional I expected to see somewhere in the code
```
if not managed:
  verify that clone opts were provided
```
otherwise you just let people with hybrid runtimes to get some weird error

rotem-codefresh: i added some checks for repo and git_token

cmd/commands/runtime.go
```diff
@@ -1762,30 +1805,16 @@ func getApplicationChecklistState(name string, a *argocdv1alpha1.Application, ru
...
-		return fmt.Errorf("failed to remove runtime dir '%s' from shared config repo. error: %w", opts.RuntimeName, err)
-	}
-
-	pushMsg := fmt.Sprintf("Removing runtime dir '%s'", opts.RuntimeName)
-	err = apu.PushWithMessage(ctx, r, pushMsg)
-	if err != nil {
-		return fmt.Errorf("failed to push to git while removing runtime '%s' dir: %w", opts.RuntimeName, err)
-	}
+	_, err = appProxyClient.AppProxyIsc().RemoveRuntimeFromIscRepo(ctx)
+	log.G(ctx).Info("Removed runtime from isc repo")
```
roi-codefresh: print log only if err != nil

roi-codefresh: preferably outside this function in the main flow

cmd/commands/runtime.go
```diff
@@ -2699,30 +2728,15 @@ func postInstallationHandler(ctx context.Context, opts *RuntimeInstallOptions, e
...
-		iscCloneOpts.Repo, err = getIscRepo(ctx)
-		handleCliStep(reporter.UninstallStepGetIscRepoFromPlatform, "Getting isc repo from platform before rollback", err, false, true)
-		if err != nil {
-			log.G(ctx).Errorf("failed to get isc repo from platform before installation rollback: %s", err.Error())
-		}
-
-		inferProviderFromRepo(opts.InsCloneOpts)
-		iscCloneOpts.Auth = opts.InsCloneOpts.Auth
-		iscCloneOpts.Progress = opts.InsCloneOpts.Progress
-		iscCloneOpts.Parse()
```
roi-codefresh, line 2715: Why do we not need this anymore?

rotem-codefresh: i've moved the isc removal to app-proxy so no need for isc clone opts

cmd/commands/runtime.go
```diff
@@ -474,12 +487,22 @@ func runtimeUpgradeCommandPreRunHandler(cmd *cobra.Command, args []string, opts
...
 	}
 
+	isManaged, err := isRuntimeManaged(cmd.Context(), opts.RuntimeName)
+	handleCliStep(reporter.UpgradeStepPreCheckIsManagedRuntime, "Checking if runtime is hosted", err, true, false)
+	if err != nil {
+		return err
+	}
+
+	if isManaged {
+		return fmt.Errorf("manual upgrades are not allowed for hosted runtimes and are managed by Codefresh operational team")
```
roi-codefresh, line 495:
```suggestion
		return fmt.Errorf("manual upgrades are not allowed for hosted runtimes as they are managed by Codefresh")
```

cmd/commands/common.go
```diff
@@ -200,10 +200,18 @@ func getRuntimeNameFromUserSelect(ctx context.Context) (string, error) {
...
 
-	runtimeNames := make([]string, len(runtimes))
+	var runtimeNames []string
 
-	for index, rt := range runtimes {
-		runtimeNames[index] = rt.Metadata.Name
+	for _, rt := range runtimes {
+		rtDisplay := rt.Metadata.Name
+		if rt.Managed {
+			if cmdFlow == RuntimeUpgrade {
```
roi-codefresh: I think instead of this whole cmdFlow type, which seems a bit overkill, you can just pass along `allowManaged bool`. You can even do it as optional args with the spread operator so it will not break existing calls, and you can just pass `true` from flows where you want to list managed runtimes.

roi-codefresh: Another good reason why not to do it like this is that the decision of which flow is allowed for managed and which flow isn't should definitely not be in `getRuntimeNameFromUserSelect`.

cmd/commands/runtime.go
rotem-codefresh: Moran wrote this log

roi-codefresh (approved):

PR #467, opened 2022-07-03
CR-12777 cluster add flow improvoments

rotem-codefresh:


cmd/commands/cluster.go
```diff
@@ -219,10 +279,16 @@ func createAddClusterKustomization(ingressUrl, contextName, server, csdpToken, v
...
 	}
 	k.FixKustomizationPostUnmarshalling()
 	util.Die(k.FixKustomizationPreMarshalling())
-	return k
+	return k, nameSuffix
+}
+
+func getClusterResourcesNameSuffix() string {
+	now := time.Now()
+	return fmt.Sprintf("%d", now.UnixMilli())
```
ghost: i think it will be simpler to just return `fmt.Sprintf("-%d", now.UnixMilli())`
(have the "-" in the suffix)

ghost: it will save all those tiny changes you made everywhere with the names of all resources

noam-codefresh (approved):

PR #485, opened 2022-07-13
CR-13046 cluster name bug

rotem-codefresh:


cmd/commands/cluster.go
```diff
@@ -200,36 +200,43 @@ func setClusterName(ctx context.Context, opts *ClusterAddOptions) error {
 }
 
 func validateClusterName(name string) error {
-	maxDNSNameLength := 253
-	if len(name) > maxDNSNameLength {
-		return fmt.Errorf("cluster name can contain no more than 253 characters")
+	maxNameLength := 63
+	if len(name) > maxNameLength {
+		return fmt.Errorf("cluster name can contain no more than 63 characters")
 	}
```
ghost, line 210: why do we need this, if we already have the length in the regexp?

cmd/commands/cluster.go
```diff
@@ -200,36 +200,43 @@ func setClusterName(ctx context.Context, opts *ClusterAddOptions) error {
...
+	if len(name) > (maxNameLength - 2) {
+		name = name[:(maxNameLength - 2)]
 	}
 
-	return strings.Trim(name, "-.")
+	name = strings.Trim(name, "-.")
+	beginsWithNum := regexp.MustCompile(`^\d`)
+	
+	if beginsWithNum.MatchString(name) {
+		name = fmt.Sprint("a", name)
```
ghost: what does that do? just add "a" at the beginning?
what if now the name is too long? why "a" and not anything else? why not just remove all digits in the prefix?
also - why not `"a"+name`?

cmd/commands/cluster.go
```diff
@@ -193,43 +193,59 @@ func setClusterName(ctx context.Context, opts *ClusterAddOptions) error {
...
+	if len(name) > (maxNameLength - 2) {
+		name = name[:(maxNameLength - 2)]
+	}
+
+	name = strings.Trim(name, "-.")
+	beginsWithNum := regexp.MustCompile(`^\d+`)
+	
+	if beginsWithNum.MatchString(name) {
+		name = beginsWithNum.ReplaceAllString(name, "")
+	}
```
ghost: what happens if you just run
```
name = beginsWithNum.ReplaceAllString(name, "")
```
wouldn't it just not do anything if there are no digits at the start?

noam-codefresh (approved):

PR #494, opened 2022-07-19
Cr 10692 support bitbucket

ghost:


manifests/app-proxy/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ kind: Kustomization
 images:
   - name: quay.io/codefresh/cap-app-proxy
     newName: quay.io/codefresh/cap-app-proxy
-    newTag: 1.1572.0
+    newTag: CR-11073
```
ghost: to be reverted

manifests/runtime.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       url: github.com/codefresh-io/cli-v2/manifests/argo-workflows
     - name: app-proxy
       type: kustomize
-      url: github.com/codefresh-io/cli-v2/manifests/app-proxy
+      url: manifests/app-proxy
```
ghost: to be reverted

pkg/git/provider_bitbucket-server.go
```diff
@@ -0,0 +1,59 @@
...
+func (bbs *bitbucketServer) Type() ProviderType {
+	return BITBUCKET_SERVER
+}
+
+func (bbs *bitbucketServer) ApiUrl() string {
+	return bbs.apiURL
+}
+
+func (bbs *bitbucketServer) VerifyToken(ctx context.Context, tokenType TokenType, token string) error {
+	return nil
```
roi-codefresh, line 57: Maybe some TODO here if you're not going to implement it now (also for gitlab)

cmd/commands/git-source.go
```diff
@@ -864,7 +864,7 @@ func createGithubExamplePipeline(opts *gitSourceGithubExampleOptions) error {
 	return nil
 }
 
-func createGithubExampleIngress(ingressClass string, ingressHost string, hostName string, ingressController ingressutil.IngressController, runtimeName string) *netv1.Ingress {
+func createGithubExampleIngress(ingressClass string, _ string, hostName string, ingressController ingressutil.IngressController, runtimeName string) *netv1.Ingress {
```
roi-codefresh: remove this param if it's not being used.

cmd/commands/common.go
```diff
@@ -279,57 +279,44 @@ func getIngressClassFromUserSelect(ingressClassNames []string) (string, error) {
...
-	if cloneOpts.Auth.Password == "" {
+	} else if cloneOpts.Auth.Password == "" {
 		return fmt.Errorf("must provide a git token using --git-token")
 	}
 
 	return nil
 }
 
-func ensureGitPAT(cmd *cobra.Command, opts *RuntimeInstallOptions) error {
+func ensureGitPAT(ctx context.Context, opts *RuntimeInstallOptions) error {
```
roi-codefresh, line 309: Can we add a comment on `ensureGitPat` and `ensureGitToken` explaining what each does, because reading it makes me wonder what is the difference and why we need both.

roi-codefresh (approved):

PR #499, opened 2022-07-23
Fix typo in "timed ot"

todaywasawesome:
Signed-off-by: Dan Garfield <dan@codefresh.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION=v0.0.443
+VERSION=v0.0.444
 
 OUT_DIR=dist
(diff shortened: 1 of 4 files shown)


noam-codefresh:
thank you for the fix, Dan.
can you please bump the version as well? just search for the version string (from the `Makefile` first line, without the `v`), everywhere in the project, and bump it by one. it should replace 4 occurrences in 3 files - `docs/releaes/release_notes.md`, `manifests/runtime.yaml` and `Makefile` itself.

roi-codefresh:
Done. LGTM.

PR #519, opened 2022-08-08
Cr 13549

danielm-codefresh:


cmd/commands/integrations.go
```diff
@@ -41,9 +41,9 @@ type (
 )
 
 var gitProvidersByName = map[string]model.GitProviders{
-	"bitbucket-server": model.GitProvidersBitbucketServer,
-	"github":           model.GitProvidersGithub,
-	"gitlab":           model.GitProvidersGitlab,
+	"BITBUCKET-SERVER": model.GitProvidersBitbucketServer,
+	"GITHUB":           model.GitProvidersGithub,
+	"GITLAB":           model.GitProvidersGitlab,
```
ghost: we can't change these.
that breaks the current cf api, and also it does not match the provider strings used by autopilot.

pkg/git/provider_bitbucket-server.go
```diff
@@ -30,7 +30,7 @@ type (
 
 const (
 	BITBUCKET_REST_ENDPOINT              = "/rest/api/1.0"
-	BITBUCKET_SERVER        ProviderType = "bitbucket-server"
+	BITBUCKET_SERVER        ProviderType = "BITBUCKET-SERVER"
```
ghost: same here, and the other 2 providers

noam-codefresh (approved):

PR #532, opened 2022-08-14
CR-13257-git-token-validation

ghost:


cmd/commands/common.go
```diff
@@ -321,7 +321,7 @@ func ensureGitPAT(ctx context.Context, opts *RuntimeInstallOptions) error {
 	}
 
 	if opts.gitProvider != nil {
-		return opts.gitProvider.VerifyToken(ctx, cfgit.PersonalToken, opts.InsCloneOpts.Auth.Password)
+		return opts.gitProvider.VerifyUserToken(ctx, opts.GitIntegrationRegistrationOpts.Token)
```
ghost, line 324: the `--personal-git-token` value ends up in `opts.GitIntegrationRegistrationOpts.Token`.
before - we checked the runtimeToken again by mistake

ghost, line 324: the `--personal-git-token` value ends up in `opts.GitIntegrationRegistrationOpts.Token`.
before - we checked the runtimeToken again by mistake

pkg/git/provider.go
```diff
@@ -21,23 +21,18 @@ import (
...
 
 	// Provider represents a git provider
 	Provider interface {
-		Type() ProviderType
 		BaseURL() string
-		VerifyToken(ctx context.Context, tokenType TokenType, token string) error
 		SupportsMarketplace() bool
+		Type() ProviderType
+		VerifyRuntimeToken(ctx context.Context, token string) error
+		VerifyUserToken(ctx context.Context, token string) error
```
ghost, line 35: split into two different funcs, instead of sending the type, to make it simpler

pkg/git/provider_bitbucket-server.go
```diff
@@ -46,22 +54,82 @@ func NewBitbucketServerProvider(baseURL string) (Provider, error) {
...
+	username, err := bbs.getCurrentUsername(ctx, token)
+	if err != nil {
+		return err
+	}
+
+	urlPath := fmt.Sprintf("users/%s/repos", username)
+	body := &createRepoBody{
+		Name: "!invalid",
+	}
+	res, err := bbs.request(ctx, token, http.MethodPost, urlPath, body)
```
ghost, line 95: POST to `users/<username>/repos` with an invalid repo name.
if it returns 400 - the token has "Project admin" permission
otherwise - the token does not have the permission

pkg/git/provider_bitbucket-server.go
```diff
@@ -46,22 +54,82 @@ func NewBitbucketServerProvider(baseURL string) (Provider, error) {
...
+		return errors.New("personal-git-token is invalid")
+	}
+
 	return nil
 }
 
-func (bbs *bitbucketServer) SupportsMarketplace() bool {
-	return false
+func (bbs *bitbucketServer) getCurrentUsername(ctx context.Context, token string) (string, error) {
+	res, err := bbs.request(ctx, token, http.MethodHead, "application-properties", nil)
```
ghost, line 123: HEAD to `application-properties` - this endpoint does not require any permission (or auth header). but any request with a valid token has the `X-AUSERNAME` response header with the user name in it.

pkg/git/provider_bitbucket-server.go
```diff
@@ -46,22 +54,82 @@ func NewBitbucketServerProvider(baseURL string) (Provider, error) {
...
+
+	if res.StatusCode != http.StatusBadRequest {
+		return errors.New("git-token is invalid or missing required \"Project admin\" scope")
+	}
+
+	return nil
+}
+
+func (bbs *bitbucketServer) checkRepoReadPermission(ctx context.Context, token string) error {
+	username, err := bbs.getCurrentUsername(ctx, token)
```
ghost: if there is no username in the response headers - that means the token was invalid.
in bitbucket-server - all tokens have "Repository read" permission. the only way to not have this permission, is to use an invalid token.

pkg/git/provider_gitlab.go
```diff
@@ -47,22 +51,64 @@ func NewGitlabProvider(baseURL string) (Provider, error) {
...
+func (g *gitlab) VerifyRuntimeToken(ctx context.Context, token string) error {
+	return g.checkApiScope(ctx, token)
+}
+
+func (g *gitlab) VerifyUserToken(ctx context.Context, token string) error {
+	return g.checkReadRepositoryScope(ctx, token)
+}
+
+func (g *gitlab) checkApiScope(ctx context.Context, token string) error {
+	res, err := g.request(ctx, token, http.MethodPost, "projects")
```
ghost, line 83: POST to `projects` without a body.
if it returns 400 - the token has "api" scope
otherwise - the token does not have the scope

pkg/git/provider_gitlab.go
```diff
@@ -47,22 +51,64 @@ func NewGitlabProvider(baseURL string) (Provider, error) {
...
+		return errors.New("git-token is invalid or missing required \"api\" scope")
+	}
+
 	return nil
 }
 
-func (g *gitlab) SupportsMarketplace() bool {
-	return false
+func (g *gitlab) checkReadRepositoryScope(ctx context.Context, token string) error {
+	res, err := g.request(ctx, token, http.MethodHead, "projects")
```
ghost, line 100: HEAD to projects.
if it returns 200 - the token has "repo_read" scope
otherwise - the token does not have the scope

pkg/git/provider_bitbucket-server.go
roi-codefresh: Please write this github comment inside the code

pkg/util/http/http.go
```diff
@@ -0,0 +1,45 @@
...
+package http
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+func Request(ctx context.Context, method, url string, headers map[string]string, body interface{}) (*http.Response, error) {
```
roi-codefresh: 👍

roi-codefresh (approved):

pkg/git/provider.go
```diff
@@ -17,50 +17,49 @@ package git
...
-var providers = map[ProviderType]func(string) (Provider, error){
+var providers = map[ProviderType]func(string, *http.Client) (Provider, error){
 	BITBUCKET_SERVER: NewBitbucketServerProvider,
 	GITHUB:           NewGithubProvider,
 	GITHUB_ENT:       NewGithubProvider, // for backward compatability
 	GITLAB:           NewGitlabProvider,
 }
 
 func GetProvider(providerType ProviderType, baseURL string) (Provider, error) {
+	client := &http.Client{}
```
ghost, line 47: this is where we can add anything we want to the client, to enable retries, or any other such settings

roi-codefresh (approved):

PR #549, opened 2022-08-28
CR-14126 support bitbucket cloud

kim-codefresh:


pkg/runtime/runtime.go
```diff
@@ -329,3 +334,32 @@ func buildFullURL(urlString string, version *semver.Version, devMode bool) strin
...
+		if err == nil {
+			break
+		}
+
+		log.G(ctx).WithFields(log.Fields{
+			"retry": try,
+			"err":   err.Error(),
+		}).Warn("Failed to create application , trying again in a second...")
+
+		time.Sleep(time.Second)
```
ghost: this didn't work without the sleep? just immediately retrying after the 1st failure?

cmd/commands/runtime_install.go
```diff
@@ -1009,12 +1012,12 @@ func installComponents(ctx context.Context, opts *RuntimeInstallOptions, rt *run
 	var err error
 
 	if !store.Get().SkipIngress && rt.Spec.IngressController != string(ingressutil.IngressControllerALB) {
-		if err = createWorkflowsIngress(ctx, opts, rt); err != nil {
+		if err = tryToCreateWorkflowsIngress(ctx, opts, rt); err != nil {
```
ghost: the function names can remain the same. functions always "try" to do something, and they might fail and return an error.
the only change is that now, inside the function, there is a retry loop.

cmd/commands/runtime_install.go
```diff
@@ -2212,3 +2172,184 @@ func initializeGitSourceCloneOpts(opts *RuntimeInstallOptions) {
...
 	opts.GsCloneOpts.Repo = host + orgRepo + "_git-source" + suffix + "/resources" + "_" + opts.RuntimeName
 }
+
+// bitbucket cloud take more time to push a commit
+// all coming retries perpuse is to avoid issues of cloning before pervious commit was pushed
+
+func persistOrUpdateRuntime(ctx context.Context, opts *RuntimeInstallOptions, rt *runtime.Runtime, server string) error {
+
+	var err error
+	for try := 0; try < pushRetries; try++ {
```
ghost: maybe we can make some `func Retry` in utils, that accepts a func, and just retries it.
it can have some extra optional parameters, like total number of retries, and optional sleep time, with sensible defaults.

pkg/runtime/runtime.go
kim-codefresh: oh I forgot, will try without!

noam-codefresh (approved):

PR #556, opened 2022-08-31
CR-14124 git integrestion step for bitbucket cloud

kim-codefresh:


cmd/commands/common.go
```diff
@@ -296,6 +297,16 @@ func ensureGitUserToken(ctx context.Context, opts *RuntimeInstallOptions) error
 		}
 
 		log.G(ctx).Infof("Personal git token was not provided. Using runtime git token to register user: \"%s\". You may replace your personal git token at any time from the UI in the user settings", currentUser.Name)
+
+		if opts.GitIntegrationRegistrationOpts.Username == "" {
```
ghost: how can the GitIntegrationRegistrationOpts.Username be anything but `""` at this point? is it connected to a new flag?

cmd/commands/integrations.go
```diff
@@ -474,7 +493,7 @@ func verifyOutputFormat(format string, allowedFormats ...string) error {
 	return fmt.Errorf("invalid output format: %s", format)
 }
 
-func parseGitProvider(provider string) (model.GitProviders, error) {
+func ParseGitProvider(provider string) (model.GitProviders, error) {
```
ghost: why make this func public? it's only being used inside this package

kim-codefresh: I was using it to verify that the username would pass for bitbucket only

noam-codefresh (approved):

PR #566, opened 2022-09-06
CR-14253-ingressless-flag

ghost:


manifests/runtime.yaml
```diff
@@ -1,11 +1,11 @@
...
 kind: Runtime
 metadata:
-  name: "{{ name }}"
-  namespace: "{{ namespace }}"
+  name: '{{ name }}'
+  namespace: '{{ namespace }}'
 spec:
-  defVersion: 1.0.1
-  version: 0.0.510
+  defVersion: 2.0.0
```
roi-codefresh, line 7: remind me what this is doing? something with upgrade right?

manifests/runtime.yaml
```diff
@@ -25,3 +25,7 @@ spec:
     - name: sealed-secrets
       type: kustomize
       url: github.com/codefresh-io/cli-v2/manifests/sealed-secrets
+    - name: frpc
+      type: helm
+      url: github.com/codefresh-io/cli-v2/manifests/frpc?ref=CR-14253-ingressless-flag
```
roi-codefresh: you are going to change that later to remove ref?

ghost: yeah, it will be completely different. i am talking to Ilia and hopefully we will change it tonight, or they will change it tomorrow morning, to the "real" repository.

manifests/runtime.yaml
ghost: yeah. all the binaries that are out there currently have `maxDefVersion = "1.0.1"`, which means that if they will encounter a `runtime.yaml` with a higher version, they will immediately fail the installation, and let the user know they should upgrade their cli-v2 binary.
user can still install using the `--version <whatever_old_version>` if they *really* want to.

but only newer binaries will know how to handle definitions with version 2.0.0 (which now include the `feature` field.

manifests/frpc/templates/deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+  selector:
+    matchLabels:
+      app: frpc
+  template:
+    metadata:
+      labels:
+        app: frpc
+    spec:
+      containers:
+        - image: quay.io/codefresh/frpc:wss1 # temp image until we have an official one
```
roi-codefresh: Do we have an official one?

manifests/frpc/templates/deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+      volumes:
+        - name: frpc-vol
+          projected:
+            sources:
+              #  - secret:
+              #      name: frpc-certs
+              #      optional: false
+              #      items:
+              #        - key: tls_trusted_ca_file
+              #          path: ca.crt
```
roi-codefresh: what's this? are we not using it?

roi-codefresh (approved):

manifests/frpc/templates/deployment.yaml
ghost: this will be enabled soon, once dev-ops finish setting up the certificate file

PR #571, opened 2022-09-14
CR-14575-support-1.24

ghost:


cmd/commands/runtime_install.go
```diff
@@ -643,12 +643,12 @@ func runRuntimeInstall(ctx context.Context, opts *RuntimeInstallOptions) error {
 	opts.RuntimeToken = token
 	opts.RuntimeStoreIV = iv
 
+	rt.Spec.IngressHost = opts.IngressHost
+	rt.Spec.InternalIngressHost = opts.InternalIngressHost
+	rt.Spec.Repo = opts.InsCloneOpts.Repo
```
ghost, line 648: related to ingressless - 
we still have an ingressHost in ingressLess (calculated from accountId, runtimeName and some suffix)
the internalIngressHost would probably still be `""`
and repo should be there anyway

roi-codefresh (approved):

PR #584, opened 2022-09-27
[CLI]:Add TCP connection check

PhilippPlotnikov:


pkg/store/store.go
```diff
@@ -236,6 +238,8 @@ func init() {
 	s.NetworkTesterName = "cf-network-tester"
 	s.NetworkTesterGenerateName = "cf-network-tester-"
 	s.NetworkTesterImage = "quay.io/codefresh/cf-venona-network-tester:latest"
+	s.TCPConnectionTesterName = "cf-tcp-connection-tester"
+	s.TCPConnectionTesterImage = "quay.io/codefresh/cf-venona-network-tester:latest"
```
ghost: why do we need a different tester image, if both the http and tcp network tests use the same one?

PhilippPlotnikov: I thought about this as in UI. We had sometimes similar logic but logically it is different entity and in future it can be changed and in code it is more clearer. I use this logic

pkg/util/kube/kube.go
```diff
@@ -61,8 +70,10 @@ type (
 	}
 )
 
-func EnsureClusterRequirements(ctx context.Context, kubeFactory kube.Factory, namespace string, contextUrl string) error {
+func EnsureClusterRequirements(runtimeInstallOptions RuntimeInstallOptions, ctx context.Context) error {
```
ghost: the `ctx context.Context` is usually the 1st func parameter wherever it is being used

pkg/store/store.go
```diff
@@ -236,6 +238,7 @@ func init() {
 	s.NetworkTesterName = "cf-network-tester"
 	s.NetworkTesterGenerateName = "cf-network-tester-"
 	s.NetworkTesterImage = "quay.io/codefresh/cf-venona-network-tester:latest"
+	s.TCPConnectionTesterName = "cf-tcp-connection-tester"
```
ghost: isn't it used as a GenerateName? i think the value should end with `-` so that it will be easier to read later
also - the `TCPConnectionTesterName` should probably be `TCPConnectionTesterGenerateName` (or maybe just `TcpTesterGenerateName` ?)

pkg/store/store.go
```diff
@@ -134,6 +134,8 @@ type Store struct {
 	NetworkTesterName                 string
 	NetworkTesterGenerateName         string
 	NetworkTesterImage                string
+	TCPConnectionTesterName           string
+	TCPConnectionTesterImage          string
```
ghost: can be removed, not used anymore (i think)

pkg/util/kube/kube.go
```diff
@@ -173,16 +184,66 @@ func EnsureClusterRequirements(ctx context.Context, kubeFactory kube.Factory, na
...
+		if err != nil {
+			return fmt.Errorf("cluster TCP connection tests failed: %w ", err)
+		}
+
+		log.G(ctx).Info("TCP connection test finished successfully")
+	}
 	return nil
 }
 
+func runTCPConnectionTest(ctx context.Context, runtimeInstallOptions *RuntimeInstallOptions) error {
```
ghost, line 206: can this func, or maybe some of this func, be merged with the func that tests http network?
or have you already extracted all the similar code from both, as much as it makes sense?

PhilippPlotnikov: I think yes. I dont see what I can move to function except handlePodState

noam-codefresh (approved):

PR #585, opened 2022-09-28
CR-14744 - fix skip-ingress flag

ghost:
fixes a regression introduced in 0.0.511

### DIFF
--- a/cmd/commands/runtime_install.go
+++ b/cmd/commands/runtime_install.go
@@ -110,6 +110,7 @@ type (
 		TunnelDomain                   string
 		TunnelSubdomain                string
 		SkipIngress                    bool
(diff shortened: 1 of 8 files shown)


cmd/commands/runtime_install.go
```diff
@@ -2045,3 +2046,7 @@ func (opts *RuntimeInstallOptions) GetValues(name string) (string, error) {
 		return "", nil
 	}
 }
+
+func (opts *RuntimeInstallOptions) InstallIngress() bool {
```
roi-codefresh:
```suggestion
func (opts *RuntimeInstallOptions) shouldInstallIngress() bool {
```

roi-codefresh (approved):

PR #593, opened 2022-10-06
CR-14554-ir

yaroslav-codefresh:


cmd/commands/runtime.go
```diff
@@ -916,6 +933,55 @@ func NewRuntimeLogsCommand() *cobra.Command {
 	return cmd
 }
 
+func migrateInternalRouter(ctx context.Context, opts *RuntimeUpgradeOptions, newRt *runtime.Runtime) error {
+	dbRuntime, err := cfConfig.NewClient().V2().Runtime().Get(ctx, opts.RuntimeName)
```
ghost: just call `getRuntime` (def is in `common.go`)

cmd/commands/runtime_install.go
```diff
@@ -20,10 +20,13 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	billyUtils "github.com/go-git/go-billy/v5/util"
```
ghost, line 23: it doesn't matter much - but why did the imports jump around like this?

cmd/commands/runtime_install.go
```diff
@@ -1036,10 +1053,33 @@ func installComponents(ctx context.Context, opts *RuntimeInstallOptions, rt *run
...
+					useGatewayAPI:             opts.useGatewayAPI,
+				}, rt)
 			},
 		}); err != nil {
-			return fmt.Errorf("failed to patch Argo-Workflows ingress: %w", err)
+			return fmt.Errorf("failed to patch Internal Router ingress: %w", err)
+		}
+	}
+
+	if opts.shouldInstallIngress() && rt.Spec.IngressController != string(routingutil.IngressControllerALB) {
```
ghost: isn't that the exact same condition as line 1053? can't it be inside the same `if`?

manifests/default-resources/internal-router/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+  - ../../internal-router
```
ghost, line 4: is this file being used anywhere?

yaroslav-codefresh: not sure I even understand the purpose of the `default-resources` folder -- just did it the same as for other components

noam-codefresh (dismissed):

noam-codefresh:
i am working on quickly fixing the upgrade regression as well (in this pr or in another), and then i'll re-approve this one.
anyway - be safe.

noam-codefresh (approved):

PR #608, opened 2022-10-25
CR-14985-flag

yaroslav-codefresh:


manifests/internal-router/internal-router.cm.yaml
```diff
@@ -16,6 +16,7 @@ data:
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
+        chunked_transfer_encoding off;
```
ziv-codefresh, line 19: is it still needed when we using version 1.22?
is so, how come that it's works out of the box up until now? maybe we are using very old version of nginx?

pkg/store/store.go
```diff
@@ -30,7 +30,7 @@ var (
 	buildDate                = ""
 	gitCommit                = ""
 	SegmentWriteKey          = ""
-	maxDefVersion            = "2.1.0"
+	maxDefVersion            = "2.1.1"
```
ziv-codefresh, line 33: what is it exactly?

ziv-codefresh (approved):

PR #620, opened 2022-11-02
move runtime definition from cli-v2 to csdp-official

danielm-codefresh:


cmd/commands/runtime_install.go
```diff
@@ -2131,5 +2129,5 @@ func getRuntimeDef(runtimeDef, versionStr string) string {
 	}
 
 	// specific version means the runtimeDef is the default value in cli-v2 repo
```
ghost: maybe just fix the comment to mention the default is in csdp-official now

Makefile
```diff
@@ -11,7 +11,7 @@ RUNTIME_DEF_URL="https://github.com/codefresh-io/cli-v2/releases/latest/download
 ADD_CLUSTER_DEF_URL="https://github.com/codefresh-io/csdp-official/add-cluster/kustomize"
 FALLBACK_ADD_CLUSTER_DEF_URL="https://github.com/codefresh-io/cli-v2/manifests/add-cluster/kustomize"
 
-DEV_RUNTIME_DEF_URL="manifests/runtime.yaml"
+DEV_RUNTIME_DEF_URL="<csdp-official-clone>/csdp/hybrid/runtime.yaml"
```
ghost: how is that supposed to work? this is the value that will be injected to the store in `make local`

pkg/store/store.go
```diff
@@ -31,11 +31,13 @@ var (
 	buildDate                = ""
 	gitCommit                = ""
 	SegmentWriteKey          = ""
-	maxDefVersion            = "2.1.1"
-	RuntimeDefURL            = "manifests/runtime.yaml"
+	maxDefVersion            = "2.1.2"
```
roi-codefresh, line 35: I don't think we need this anymore

pkg/store/store.go
```diff
@@ -151,16 +155,33 @@ type Store struct {
 	IngressHost                       string
 	IscRuntimesDir                    string
 	DevMode                           bool
+	DefVersionComptability            map[string]string
```
roi-codefresh:
```suggestion
	DefVersionToLastCLIVersion            map[string]string
```

pkg/store/store.go
```diff
@@ -151,16 +155,33 @@ type Store struct {
...
+	if versionStr == "" {
+		return runtimeDefURL
+	}
+
+	version := semver.MustParse(versionStr)
+	if version.Compare(s.LastRuntimeVersionInCLI) <= 0 {
+		runtimeDefURL = OldRuntimeDefURL
+	}
+
+	return runtimeDefURL
```
roi-codefresh: let's to where it is called

cmd/commands/runtime.go
```diff
@@ -817,6 +817,8 @@ func NewRuntimeUpgradeCommand() *cobra.Command {
 				finalParameters["Version"] = versionStr
```
roi-codefresh: remove `versionStr` var, looks dead

roi-codefresh: replace with opts.versionStr

cmd/commands/runtime.go
```diff
@@ -878,6 +880,14 @@ func runRuntimeUpgrade(ctx context.Context, opts *RuntimeUpgradeOptions) error {
 		return err
 	}
 
+	requiredCliVersionConstraint, _ := semver.NewConstraint(newRt.Spec.RequiredCLIVersion)
```
roi-codefresh: explain why ignoring the err here, what are the possible flows here

cmd/commands/runtime.go
```diff
@@ -878,6 +880,14 @@ func runRuntimeUpgrade(ctx context.Context, opts *RuntimeUpgradeOptions) error {
...
+	if requiredCliVersionConstraint != nil && !requiredCliVersionConstraint.Check(store.Get().Version.Version) {
+		err = fmt.Errorf("to upgrade to this version, please install cli version %s", requiredCliVersionConstraint.String())
```
roi-codefresh:
```suggestion
		return fmt.Errorf("to upgrade to this version, please install cli version %s", requiredCliVersionConstraint.String())
```

cmd/commands/runtime_install.go
```diff
@@ -222,6 +222,8 @@ func NewRuntimeInstallCommand() *cobra.Command {
 				return util.DecorateErrorWithDocsLink(fmt.Errorf("pre installation error: %w", err), store.Get().RequirementsLink)
 			}
 
+			installationOpts.runtimeDef = store.GetRuntimeDefURL(installationOpts.versionStr)
```
roi-codefresh: need to check if this is assigned before overwriting

roi-codefresh: for upgrade also

cmd/commands/runtime_install.go
```diff
@@ -1143,6 +1146,14 @@ func preInstallationChecks(ctx context.Context, opts *RuntimeInstallOptions) (*r
...
 		err = fmt.Errorf("your cli version is out of date. please upgrade to the latest version before installing")
+	} else if rt.Spec.DefVersion.LessThan(store.Get().MaxDefVersion) {
+		val := store.Get().DefVersionComptability[rt.Spec.DefVersion.String()]
+		err = fmt.Errorf("to install this version, please downgrade your cli to version %s", val)
+	}
+
+	requiredCliVersionConstraint, _ := semver.NewConstraint(rt.Spec.RequiredCLIVersion)
+	if requiredCliVersionConstraint != nil && requiredCliVersionConstraint.Check(store.Get().Version.Version) {
+		err = fmt.Errorf("to install this version, please install cli version %s", requiredCliVersionConstraint.String())
 	}
```
roi-codefresh: please extract this operation to runtime.go pkg

cmd/commands/runtime_install.go
```diff
@@ -1143,6 +1146,14 @@ func preInstallationChecks(ctx context.Context, opts *RuntimeInstallOptions) (*r
 
 	if rt.Spec.DefVersion.GreaterThan(store.Get().MaxDefVersion) {
 		err = fmt.Errorf("your cli version is out of date. please upgrade to the latest version before installing")
+	} else if rt.Spec.DefVersion.LessThan(store.Get().MaxDefVersion) {
```
roi-codefresh: If defVersion is nil on the runtime spec you should ignore it

cmd/commands/runtime_install.go
```diff
@@ -2135,6 +2146,6 @@ func getRuntimeDef(runtimeDef, versionStr string) string {
 		return runtimeDef
 	}
 
-	// specific version means the runtimeDef is the default value in cli-v2 repo
-	return strings.Replace(runtimeDef, "/releases/latest/download", "/releases/download/v"+version.String(), 1)
+	// specific version means the runtimeDef is the default value in cli-v2/csdp-official repo
+	return runtimeDef + "?ref=v" + version.String()
```
roi-codefresh: check if new or old runtime

pkg/store/store.go
roi-codefresh, line 35: please do not touch this field it is deprecated, it's only here to allow to install runtimes with version < xyz

roi-codefresh (approved):

PR #630, opened 2022-11-06
replace warning to error log

kim-codefresh:


roi-codefresh (approved):

manifests/runtime.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: "{{ namespace }}"
 spec:
   defVersion: 2.1.1
-  version: 0.0.569
+  version: 0.0.570
```
danielm-codefresh: No need to bump this version anymore.
This file is deprecated and is only here for backward compatibility.

danielm-codefresh: Actually can you please add a comment at the top of the file saying that?

danielm-codefresh (approved):

manifests/runtime.yaml
kim-codefresh: yes

PR #645, opened 2022-11-29
Create /home/codefresh

lrochette:
When using the image, you need to create a context By default the context is created in /home/codefresh but this folder does not exist It would be a lot easier if the folder exists instead of having to create and manage the context inna different locations


lrochette:
@noam-codefresh Is this PR still required. It seems it was merged in a different patch

PR #646, opened 2022-11-30
Revise CLI wizard to support access mode and git provider

danielm-codefresh:


cmd/commands/common.go
```diff
@@ -323,7 +325,7 @@ func ensureGitUserToken(ctx context.Context, opts *RuntimeInstallOptions) error
 
 func getGitTokenFromUserInput(cmd *cobra.Command) error {
 	gitTokenPrompt := promptui.Prompt{
-		Label: "Runtime git api token",
+		Label: fmt.Sprintf("Runtime git token: (required scopes should be as described in: %s )", store.Get().GitTokensLink),
```
roi-codefresh, line 328:
```suggestion
		Label: fmt.Sprintf("Runtime git token: (required scopes should be as described in: %s)", store.Get().GitTokensLink),
```

roi-codefresh (approved):

PR #648, opened 2022-12-04
Add validation to block usage of --ingress-host flag together with --…

danielm-codefresh:
…access-mode=tunnel

### DIFF
--- a/cmd/commands/runtime_install.go
+++ b/cmd/commands/runtime_install.go
@@ -197,11 +197,12 @@ func NewRuntimeInstallCommand() *cobra.Command {
 
 			if accessMode != "" {
 				installationOpts.AccessMode = platmodel.AccessMode(strings.ToUpper(accessMode))
-				if installationOpts.AccessMode == platmodel.AccessModeTunnel && installationOpts.IngressHost != "" {
-					return fmt.Errorf("ingress host can't be set when access mode is Tunnel.")
-				}
 				if !installationOpts.AccessMode.IsValid() {
 					return fmt.Errorf("invalid access-mode %s, must be one of: ingress|tunnel", accessMode)
+				}
+
+				if installationOpts.AccessMode == platmodel.AccessModeTunnel && installationOpts.IngressHost != "" {
+					return fmt.Errorf("ingress host can't be set when access mode is Tunnel")
 				}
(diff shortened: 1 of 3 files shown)


cmd/commands/runtime_install.go
```diff
@@ -196,8 +195,11 @@ func NewRuntimeInstallCommand() *cobra.Command {
 
 			createAnalyticsReporter(ctx, reporter.InstallFlow, installationOpts.DisableTelemetry)
 
-			if (accessMode != "") {
+			if accessMode != "" {
 				installationOpts.AccessMode = platmodel.AccessMode(strings.ToUpper(accessMode))
+				if installationOpts.AccessMode == platmodel.AccessModeTunnel && installationOpts.IngressHost != "" {
+					return fmt.Errorf("ingress host can't be set when access mode is Tunnel.")
+				}
```
ghost: maybe this should be after the next condition? make sure accessMode is valid first?

cmd/commands/runtime_install.go
```diff
@@ -1145,7 +1146,11 @@ func preInstallationChecks(ctx context.Context, opts *RuntimeInstallOptions) (*r
 		return nil, err
 	}
 
-	rt := opts.DownloadRuntimeDef
+	runtimeDef := getRuntimeDef(opts.runtimeDef, opts.versionStr)
+	rt, err := runtime.Download(runtimeDef, opts.RuntimeName, opts.featuresToInstall)
+	if err != nil {
+		return nil, fmt.Errorf("failed to download runtime definition: %w", err)
+	}
```
ghost, line 1154: it doesn't make sense to keep the downloaded runtime definition on the opts? i think we can use it later. but if it is not actually being used, i guess there is no point in it

danielm-codefresh: I added this in the last version, where I moved runtime definition download to the pre command handler (where we can early check if CLI is compatible with runtime version), but now I noticed that we later check for features such as ingressless...

cmd/commands/runtime_install.go
danielm-codefresh: done

noam-codefresh (approved):

PR #654, opened 2022-12-08
CR-15158

eti-codefresh:


daniel-codefresh (approved):

roi-codefresh:
I dont like this command name, not according to convention

PR #659, opened 2022-12-19
check kube error, return link to article if needed

ghost:


pkg/util/kube/kube.go
```diff
@@ -645,3 +647,22 @@ func GetSecretsWithLabel(ctx context.Context, kubeFactory kube.Factory, namespac
 
 	return secrets, nil
 }
+
+func GetClientSet(kubeFactory apkube.Factory) (kubernetes.Interface, error) {
```
ghost, line 651: all calls will go through here

pkg/util/kube/kube.go
```diff
@@ -645,3 +647,22 @@ func GetSecretsWithLabel(ctx context.Context, kubeFactory kube.Factory, namespac
...
+	cs, err := kubeFactory.KubernetesClientSet()
+	if err != nil {
+		if strings.Contains(err.Error(), "exec plugin: invalid apiVersion") {
+			return nil, errors.New("Kubeconfig user entry is using an invalid API version client.authentication.k8s.io/v1alpha1.\nSee details at https://support.codefresh.io/hc/en-us/articles/6947789386652-Failure-to-perform-actions-on-your-selected-Kubernetes-context")
```
ghost, line 655: the link is for classic support page, but it is general enough. i don't think we need to copy it into v2 docs. wdyt?

danielm-codefresh (approved):

danielm-codefresh:
LGTM!

PR #666, opened 2023-01-08
CR-15777

kim-codefresh:


pkg/git/provider_gitlab.go
```diff
@@ -94,6 +112,36 @@ func (g *gitlab) checkApiScope(ctx context.Context, token string) error {
...
+	if err != nil {
+		return "", fmt.Errorf("failed getting user: %w", err)
+	}
+
+	bodyBytes, err := io.ReadAll(userRes.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed reading user body: %w", err)
+	}
+
+	var body useResBody
```
roi-codefresh:
```suggestion
	var user useResBody
```

pkg/git/provider_gitlab.go
```diff
@@ -32,6 +35,11 @@ type (
 		apiURL       *url.URL
 		c            *http.Client
 	}
+
+	useResBody struct {
```
roi-codefresh:
```suggestion
	gitlabUserResponse struct {
```

pkg/git/provider_gitlab.go
```diff
@@ -94,6 +112,36 @@ func (g *gitlab) checkApiScope(ctx context.Context, token string) error {
...
+		return "", fmt.Errorf("failed parse user body: %w", err)
+	}
+	if body.Bot {
+		if strings.HasPrefix(body.Username, "project") {
+			return "project", nil
+		}
+		return "group", nil
+	}
+
+	defer userRes.Body.Close()
```
roi-codefresh: Should be moved to line 121, right after err is checked. otherwise you will have a leak

roi-codefresh (approved):

PR #668, opened 2023-01-22
warn and continue in argocd login error

ghost:


noam-codefresh:
will not work, since argocd code exists (`Fatal`) on that error.
ended up making a special binary for the customer instead (that just skips `argocd login` call)

PR #671, opened 2023-02-05
 add skip-permissions-validation flag

kim-codefresh:
…token

### DIFF
--- a/cmd/commands/runtime.go
+++ b/cmd/commands/runtime.go
@@ -65,8 +65,8 @@ type (
 		DisableTelemetry bool
 		Managed          bool
 
-		kubeContext               string
-		skipAutopilotUninstall    bool
+		kubeContext            string
+		skipAutopilotUninstall bool
 	}
 
 	RuntimeUpgradeOptions struct {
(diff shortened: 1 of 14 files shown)


cmd/commands/runtime.go
```diff
@@ -187,7 +187,8 @@ func runtimeUninstallCommandPreRunHandler(cmd *cobra.Command, args []string, opt
 	}
 
 	if !opts.Managed {
-		err = ensureGitRuntimeToken(cmd, nil, opts.CloneOpts)
+		log.G(ctx).Debug("skipping permmission validation")
```
ghost: i don't think we need this log in uninstall. i think the fact we originally were checking the token permissions when we shouldn't have is the mistake.

cmd/commands/runtime.go
```diff
@@ -230,7 +231,8 @@ func runtimeUpgradeCommandPreRunHandler(cmd *cobra.Command, args []string, opts
 		return err
 	}
 
-	err = ensureGitRuntimeToken(cmd, nil, opts.CloneOpts)
+	log.G(ctx).Debug("skipping permmission validation")
```
ghost: same here

noam-codefresh (approved):

PR #692, opened 2023-05-01
Chore: Add Issue (and PR) Templates

korenyoni:
## What

* Add PR template
* Add issue template

## Why

* Improve repo hygiene


noam-codefresh (approved):

noam-codefresh:
i assume the build fails because there is no change to the version in the makefile. but there is also no real change to any actual code files. so i'll just force-merge it, and we will start using those templates and see how they work for us.
thanks @korenyoni .

PR #697, opened 2023-06-06
CR-18322-validate-values

ghost:
## What
added more helm validations (support platform and git certificates, allow ingress sklip, etc)

## Why
going to add as a pre-install step to helm-install

## Notes
changes will also be effective as stand-alone helm validation checks


pkg/config/config.go
```diff
@@ -187,36 +186,28 @@ func (c *ConfigImpl) GetCurrentContext() *AuthContext {
...
-
-	// validate new context
-	client := c.clientForContext(authCtx)
-	_, err := client.Users().GetCurrent(ctx)
-	if err != nil {
-		if url == store.Get().DefaultAPI {
-			err = fmt.Errorf("failed to create codefresh client with token \"%s\": %w", token, err)
-		} else {
-			err = fmt.Errorf("failed to create codefresh client with url \"%s\" and token \"%s\": %w", url, token, err)
-		}
```
ghost, line 214: removed check on client creation, because now we can create a client with a runtimeToken (which fails on `GetCurrent` calls)

pkg/config/config.go
```diff
@@ -248,26 +239,30 @@ func (c *ConfigImpl) UseContext(ctx context.Context, name string) error {
...
 	if _, exists := c.Contexts[name]; exists {
 		return fmt.Errorf("authentication context with the name \"%s\" already exists", name)
 	}
 
 	authCtx := &AuthContext{
-		Name:  name,
-		URL:   url,
-		Token: token,
-		Type:  "APIKey",
-		Beta:  false,
```
ghost, line 261: `Beta` was always `false` (also in codefresh cli v1, if i am not mistaked), and was also never actually used for anything.

danielm-codefresh (commented):
Check if this fails when no git credentials are provided

danielm-codefresh (approved):

PR #699, opened 2023-06-15
check account admin correctly

ghost:
## What
fixed helm validation check for user admin role. old code was checking for codefresh-admin, and not account-admin, incorrectly.

## Why
the previous helm validation check would fail for all of our customers.

## Notes
after this is merged and released - need to update the gitops-helm-chart to use the new release.


noam-codefresh:
renamed branch and pr

PR #748, opened 2024-04-16
fix(dockerfile): take `GITHUB_TOKEN` from secrets

masontikhonov:
## What

This takes `GITHUB_TOKEN` from secrets instead of build args.

## Why

Security reasons.

## Notes

—

### DIFF
--- a/build/ci.yaml
+++ b/build/ci.yaml
@@ -36,6 +36,15 @@ steps:
           on:
             - success
 
+  prepare_build_secrets:
+    stage: Prepare
+    title: 'Prepare build secrets'
+    image: alpine
+    commands:
+      - export GITHUB_TOKEN_PATH=${{CF_VOLUME_PATH}}/GITHUB_TOKEN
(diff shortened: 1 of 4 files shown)


zarbis (approved):

danielm-codefresh (approved):

build/ci.yaml
```diff
@@ -36,6 +36,15 @@ steps:
           on:
             - success
 
+  prepare_build_secrets:
+    stage: Prepare
+    title: 'Prepare build secrets'
+    image: alpine
+    commands:
+      - export GITHUB_TOKEN_PATH=${{CF_VOLUME_PATH}}/GITHUB_TOKEN
+      - cf_export GITHUB_TOKEN_PATH
```
zarbis, line 45: `--mask`?

masontikhonov: It's the path to the token, which we're constructing one line above, I don't think we should really mask it.

PR #758, opened 2024-10-22
feat: update cli to retrieve promotion commits for product release steps

scme0:
## What
Adding promotion commits to product release steps if the version of codefresh is compatible. I've added a fall back in case the cli is connected to an old version of codefresh (on prem).

## Why
A customer wanted to know the commit created by promotions for each step of a product release for each application. See Jira ticket for details.

## Notes
Jira: https://codefresh-io.atlassian.net/browse/CR-25720

### DIFF
--- a/cmd/commands/product-release.go
+++ b/cmd/commands/product-release.go
@@ -51,7 +51,7 @@ query getProductReleasesList(
 			steps {
 				environmentName
 				status
-				promotionCommits {
+				applications {
 					applicationId {
 						runtime
 						namespace
--- a/cmd/commands/product-release.go
+++ b/cmd/commands/product-release.go
@@ -67,7 +67,7 @@ query getProductReleasesList(
 }
 `
 
-const v0_13_4_query = `
+const pre_v1_3120_1_query = `
 query getProductReleasesList(
 	$productName: String!
 	$filters: ProductReleaseFiltersArgs!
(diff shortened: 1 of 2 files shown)


cmd/commands/product-release.go
```diff
@@ -126,9 +158,15 @@ query getProductReleasesList(
 		},
 	}
 
-	productReleasesPage, err := client.GraphqlAPI[productReleaseSlice](ctx, cfConfig.NewClient().InternalClient(), query, variables)
+	productReleasesPage, err := client.GraphqlAPI[productReleaseSlice](ctx, cfConfig.NewClient().InternalClient(), latest_query, variables)
 	if err != nil {
-		return fmt.Errorf("failed to get product releases: %w", err)
+		if strings.Contains(err.Error(), "Cannot query field \\\"promotionCommits\\\" on type \\\"ProductReleaseStep\\\".") {
+			log.G().Warn("codefresh version v0.13.4 or older detected. Using v0.13.4 query which excludes promotionCommits.")
```
scme0: I added this warning, but now I'm not sure as if there are customers/internal users using this cli command in automation the extra log line of output would make it fail.

I can also just make it fallback silently?

cmd/commands/product-release.go
```diff
@@ -37,6 +38,55 @@ type (
...
+				}
+			}
+			status
+			}
+		}
+	}
+}
+`
+
+const v0_13_4_query = `
```
scme0: I'll update the specific version references once I've merged the argo-platform changes.

cmd/commands/product-release.go
ATGardner: i think that should be ok, showing a warning, but falling back to previous query.
this way they will know that something may work better if they upgrade their platform.

scme0:
I've tested this against g.codefresh while it runs a pre v1.3120.1 version and it correctly logs a warning and displays the results without the applications property.
I've also tested it against sandbox running a branch build of v1.3120.1 and it successfully returns the applications property.

ATGardner (approved):

PR #780, opened 2025-04-07
Fix-release-version

ATGardner:
## What
fixed the build script to replace values under the internal package

## Why
values were not being injected to the correct place

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -9,12 +9,10 @@ fi
 
 go build -ldflags=" \
     -extldflags '-static' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.binaryName=${BINARY_NAME}' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.version=${VERSION}' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.buildDate=${BUILD_DATE}' \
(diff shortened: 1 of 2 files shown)


hack/build.sh
```diff
@@ -9,12 +9,10 @@ fi
 
 go build -ldflags=" \
     -extldflags '-static' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.binaryName=${BINARY_NAME}' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.version=${VERSION}' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.buildDate=${BUILD_DATE}' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.gitCommit=${GIT_COMMIT}' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.RuntimeDefURL=${RUNTIME_DEF_URL}' \
```
ATGardner, line 16: RuntimeDefURL is not being used anywhere

hack/build.sh
```diff
@@ -9,12 +9,10 @@ fi
 
 go build -ldflags=" \
     -extldflags '-static' \
-    -X 'github.com/codefresh-io/cli-v2/pkg/store.binaryName=${BINARY_NAME}' \
```
ATGardner, line 12: change from `pkg` to `internal`

dmaizel (approved):

PR #785, opened 2025-04-15
feat: cf account validate-limits command

andrii-codefresh:


cmd/commands/cluster.go
```diff
@@ -184,12 +184,7 @@ func runClusterAdd(ctx context.Context, opts *ClusterAddOptions) error {
 	log.G(ctx).Info("Building \"add-cluster\" manifests")
 
 	csdpToken := cfConfig.GetCurrentContext().Token
-	addClusterRef := version.String()
-	if runtime.InstallationType != platmodel.InstallationTypeCli && runtime.InstallationType != platmodel.InstallationTypeHosted {
```
dmelnyk-octo, line 188: Why it was removed? Can you elaborate more, please?

ATGardner (approved):

cmd/commands/account.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	if !limitsStatus.Status {
+		return fmt.Errorf("account limits exceeded for account: %s", string(statusString))
+	}
+
+	if opts.hook &&
```
dmelnyk-octo: I thought a bit more about the namings.
* I think better name would be `validate-usage` because "Usage" is what user see on the subscription and billing pages.
* Since CLI is the public tool I think we shoudn't put there any command and options that will be used and will be understable only by us. 
I believe we can make interface more user friendly if we use flags `fail-condition` and `subject` to control the behavior. 
```bash
--fail-condition=['reached' | 'exceeded'(default)] --subjects=['applications' | 'clusters']
```

And our command that will be used in the helm hook will be :
```bash
cf account validate-usage --fail-condition=reached --subject=clusters
```

cmd/commands/account.go
```diff
@@ -0,0 +1,105 @@
...
+		limitsStatus.Limits.Clusters != nil &&
+		limitsStatus.Usage.Clusters != nil &&
+		*limitsStatus.Limits.Clusters == *limitsStatus.Usage.Clusters {
+		limitsStatus.Status = false
+		statusString, _ := json.MarshalIndent(limitsStatus, "", "  ")
+		return fmt.Errorf("account limits (clusters) exceeded for account: %s", string(statusString))
+	}
+
+	log.G(ctx).Infof("Successfully validated limits for account")
```
dmelnyk-octo: If the command is doing validation, why do we not throw an error if the usage was exceeded?

cmd/commands/cluster.go
andrii-codefresh: The installation types used in this block have been removed from the platform. We don't need to take them into account.

cmd/commands/account.go
andrii-codefresh: Errorf creates an `error` object.

dmelnyk-octo: But when we run not in "hook" mode, we still want to fail if the usage is exceeded. Also it should be the default behavior when `failure-condition` is not provided (default is exceeded).

andrii-codefresh: We don't need hook mode anymore if we use --fail-condition=reached
It is used for the same purpose

dmelnyk-octo (approved):

cmd/commands/account.go
```diff
@@ -163,3 +198,53 @@ func ValidateGitOpsUsage(usage platmodel.GitOpsUsage, limits platmodel.GitOpsLim
 	// No validation errors
 	return nil
 }
+
+func CreatePlatformClientInRuntime(ctx context.Context, opts *ValidateLimitsOptions) (codefresh.Codefresh, error) {
```
dmelnyk-octo: Can you check if there is a better place for this function in cli-v2? It can be used by other commands in the future.

docs/releases/release_notes.md
```diff
@@ -15,7 +15,7 @@ cf version
 
 ```bash
 # download and extract the binary
-curl -L --output - https://github.com/codefresh-io/cli-v2/releases/download/v0.2.1/cf-linux-amd64.tar.gz | tar zx
+curl -L --output - https://github.com/codefresh-io/cli-v2/releases/download/v0.2.6/cf-linux-amd64.tar.gz | tar zx
```
dmelnyk-octo: 0.2.7 in Makefile

docs/releases/release_notes.md
```diff
@@ -28,7 +28,7 @@ cf version
 
 ```bash
 # download and extract the binary
-curl -L --output - https://github.com/codefresh-io/cli-v2/releases/download/v0.2.1/cf-darwin-amd64.tar.gz | tar zx
+curl -L --output - https://github.com/codefresh-io/cli-v2/releases/download/v0.2.6/cf-darwin-amd64.tar.gz | tar zx
```
dmelnyk-octo: 0.2.7 in Makefile

dmelnyk-octo (approved):

PR #793, opened 2025-08-13
chore(CR-30344): updated helm, golang.org/x/net, cloudflare/circl

vitalii-codefresh:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24.4-alpine3.22 AS base
+FROM golang:1.24.5-alpine3.22 AS base
```
ATGardner: you can bump to golang 1.24.6

ATGardner (approved):

Dockerfile
vitalii-codefresh: done

PR #801, opened 2025-12-22
fix: bitbucket server and cloud provider selection

kim-codefresh:

### DIFF
--- a/internal/git/provider_bitbucket.go
+++ b/internal/git/provider_bitbucket.go
@@ -37,7 +37,8 @@ type (
 
 const (
 	BITBUCKET_CLOUD_DOMAIN               = "bitbucket.org"
-	BITBUCKET_REST_ENDPOINT              = "/api/2.0"
+	BITBUCKET_CLOUD_API_URL              = "api.bitbucket.org"
+	BITBUCKET_REST_ENDPOINT              = "/2.0"
 	BITBUCKET               ProviderType = "bitbucket"
 )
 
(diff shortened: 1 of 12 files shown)


internal/git/provider_bitbucket.go
```diff
@@ -68,7 +69,10 @@ func NewBitbucketProvider(baseURL string, client *http.Client) (Provider, error)
 		return nil, fmt.Errorf("wrong domain for bitbucket provider: \"%s\", expected \"%s\"\n  maybe you meant to use \"bitbucket-server\" for on-prem git provider?", baseURL, BITBUCKET_CLOUD_DOMAIN)
 	}
 
+	// new api token working only with api.bitbucket.org/2.0
+	u.Host = BITBUCKET_CLOUD_API_URL
 	u.Path = BITBUCKET_REST_ENDPOINT
+	u.User = nil // make sure to remove prefix of https://user:pass@..
 	return &bitbucket{
```
ATGardner, line 77: why are we using the `baseURL` at all in here? we *know* the api url should just be "https://api.bitbucket.org/2.0". we can just ignore the `baseURL` argument completely.

internal/git/provider_gitlab.go
```diff
@@ -45,15 +45,18 @@ const (
...
 func NewGitlabProvider(baseURL string, client *http.Client) (Provider, error) {
-	u, err := url.Parse(baseURL)
+	u, err := url.Parse(baseURL) // baseURL is the isc url
 	if err != nil {
 		return nil, err
 	}
 
-	if u.Host == GITLAB_CLOUD_DOMAIN || u.Path == "" {
+	if strings.Contains(u.Hostname(), GITLAB_CLOUD_DOMAIN) {
+		u, _ = url.Parse(GITLAB_CLOUD_API_URL)
```
ATGardner, line 58: if it is in gitlab.com, shouldn't `u` end up with `https://gitlab.com/api/v4`?

kim-codefresh: gitlab is both for gitlab cloud and on prem

ATGardner (approved):

PR #803, opened 2026-02-12
feat: update argocd to v3 , update go to 1.25

pysarenko-bohdan:


cmd/commands/upgrade.go
```diff
@@ -39,7 +39,7 @@ func newUpgradeCommand() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&opts.version, "version", "", "Specify a cli version to upgrade to")
-	cmd.Flags().StringVarP(&opts.output, "ouput", "o", "", "Where to save the new binary (default: replace the old binary)")
+	cmd.Flags().StringVarP(&opts.output, "output", "o", "", "Where to save the new binary (default: replace the old binary)")
```
masontikhonov, line 42: @pysarenko-bohdan this would be a breaking change. We should either bump major version, or revert this typo 🥹
I would probably keep fix and bump major

masontikhonov (changes requested):
LGTM except for breaking change comment

masontikhonov (approved):

PR #805, opened 2026-03-31
fix: security

andrii-codefresh:
CVE-2026-33186
[CVE-2026-24051](https://osv.dev/vulnerability/GHSA-9h8m-3fm2-qjrq) - go.opentelemetry.io/otel/sdk


Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION=v1.0.1
+VERSION=v1.1.0
```
masontikhonov: @andrii-codefresh did we add new features? If not, it should be patch version according [to the spec](https://semver.org/).
```suggestion
VERSION=v1.0.2
```

masontikhonov (approved):
Left a comment, other than that looks good to me

